Issue #7, opened 2021-10-28
Check Bootstrap changes


RobertGemmaJr:
Package uses the newest versions of `bootstrap` and `react-bootstrap`. There will likely be styling issues, should go through the code with a fine toothed comb

Issue #6, opened 2021-10-26
Add README


RobertGemmaJr:

RobertGemmaJr:
Write both README's - the library and the example project. Include some screenshots of the example

Issue #9, opened 2021-11-02
Display spinner when scene is loading


RobertGemmaJr:
Spinner should be displayed over the Aframe scene in some obvious way while the model is being loaded. Code below is what was in the RIDDC project, not what the final functionality should look like.

```jsx
      <div id="modelLoaded" style={{ display: "block" }}>
        <Spinner animation="border">
          <span className="visually-hidden">Loading Volume</span>
        </Spinner>
      </div>
```

Issue #5, opened 2021-10-26
Remove aframe-react Dependency & Upgrade aframe


RobertGemmaJr:
[aframe-react](https://github.com/supermedium/aframe-react) creates a lot of memory garbage causing the website to lag. The package itself no longer recommends using it.

Remove and use "A-Frame and aframe-state-component with static templating" to display volumes

RobertGemmaJr:
Use [aframe](https://www.npmjs.com/package/aframe) dependency - not the forked version we're using now

RobertGemmaJr:
It looks like `aframe-react` is handling the model's position under the hood? Not entirely sure but if we can expose it then the reset button can revert back to the original model when clicked

PR #10, opened 2021-11-02
Initial Deployment with skeleton components

RobertGemmaJr:
Initial deployment of the library and example project.

I would appreciate feedback/discussion about the expected/default/required props for the library. I have what I think is pretty reasonable but it's definitely just a first pass. I'm sure it will be cleaned up as the project goes along. Is there a way to add prop descriptions in the code and have it auto-populate the README? Something like doxygen for npm modules.

The same is true for how we should control state once we're in the project. I copied over the context from RIDDC but looking at it now I'm not entirely sure what the best way to handle state inside the library is. I know we can use hooks to run code whenever a specific prop changes - do we need context then?

Couple things to note:
- I carried over the SASS files from RIDDC but am using styled components - left the file in for right now to make sure I add the styles that were originally created in RIDDC
- Left the aframe `<Entity />` that actually loads the model out of `AframeScene.js` for right now. It causes errors if not receive thing right data, which needs to be configured with the projects state first

- Closes issue #1 
- Closes issue #6 
- Closes issue #7


broarr:
> I would appreciate feedback/discussion about the expected/default/required props for the library. I have what I think is pretty reasonable but it's definitely just a first pass. I'm sure it will be cleaned up as the project goes along. Is there a way to add prop descriptions in the code and have it auto-populate the README? Something like doxygen for npm modules.

Seems like `react-docgen` is the most popular tool for this sort of thing.
https://www.npmtrends.com/parse-prop-types-vs-proptypes-parser-vs-react-docgen-vs-react-prop-types-parser



> The same is true for how we should control state once we're in the project. I copied over the context from RIDDC but looking at it now I'm not entirely sure what the best way to handle state inside the library is. I know we can use hooks to run code whenever a specific prop changes - do we need context then?

If this is intended to be a library I would remove the context and rely on prop changes. Contexts are relatively benign, in SOMA we use a couple of projects that provide contexts for us to consume (mostly for dark/light mode), but if we can get away without it I think it's cleaner

src/Aframe/arcball-camera.js
```diff
@@ -0,0 +1,122 @@
...
+      aCanvas.style.cursor = "grab";
+    });
+  },
+
+  onWindowResize() {},
+
+  bindMethods: function () {
+    this.onEnterVR = bind(this.onEnterVR, this);
+    this.onExitVR = bind(this.onExitVR, this);
+  },
```
broarr, line 57: This confuses me. My classical JS is out of date, but I would think that `onEnterVR` would already be bound to `this` correctly as it's defined in the same `function` declaration...

src/Aframe/arcball-camera.js
```diff
@@ -0,0 +1,122 @@
...
+
+  update: function (oldData) {
+    let controls = this.controls;
+    controls.rotateSpeed = 1.0;
+    controls.zoomSpeed = 1.2;
+    controls.panSpeed = 0.8;
+  },
+
+  onExitVR: function () {
+    console.log("exit VR");
```
broarr, line 85: For libraries I try to remove `console.log` functions, or add a logger that can be controlled by the user

broarr: Obviously that can come later :-)

src/components/App.jsx
```diff
@@ -0,0 +1,56 @@
...
+  //       path: path,
+  //       slices: slices,
+  //       spacing: spacing,
+  //       position: position,
+  //       rotation: rotation,
+  //       scale: scale,
+  //       dataRange: dataRange,
+  //     },
+  //   });
+  // }, []);
```
broarr, line 48: FYI, the `[]` dependency array means this will only fire once when the component is mounted

src/components/controls/Controls.jsx
```diff
@@ -0,0 +1,32 @@
...
+import styled from "styled-components";
+import { Stack } from "react-bootstrap";
+
+// import OpacityControls from "./OpacityControl";
+// import ColorMapControls from "./ColorMapControls";
+// import ClipControls from "./ClipControls";
+
+const ControlsStack = styled(Stack)`
+  position: absolute;
+  z-index: 1;
```
broarr, line 11: Did you see the section in CSS for JS about letting the DOM order decide the z-index? Might be a better solution here to prevent any CSS conflicts with the consumers of this library

src/components/App.jsx
```diff
@@ -0,0 +1,56 @@
...
+    path,
+    position,
+    rotation,
+    scale,
+    slices,
+    sliderRange,
+    spacing,
+    useDefaultColorMaps,
+    useTransferFunction,
+  } = props;
```
broarr, line 29: Why pull all these out if you're not using them?

broarr: Ahh, I think I get it, this is all for a skeleton deployment to make sure stuff works properly

broarr (approved):
Looks like a fine initial deployment to me!

src/components/controls/Controls.jsx
RobertGemmaJr: Yeah it was being a little finicky when I tried to do it that way. I'll have to go back and make sure all the relative/absolute positioning is correct

src/components/App.jsx
RobertGemmaJr: How would I get it to fire only when the prop is changed? When I added `path`, `slices`, etc. to the dependency array the effect ended up firing constantly

src/Aframe/arcball-camera.js
RobertGemmaJr: I'll throw it in issue #11

src/components/controls/Controls.jsx
mcmcgrath13: I think it would be fine to use z-index as long as we wrap our whole library in `isolation: isolate`

src/components/App.jsx
broarr: By default react components will rerender when their props change. I don't think you need a `useEffect` hook for that. `useEffect` is for side effects that aren't surfaced to the user via props. Try that and let me know how it goes, I might be misremembering something

src/components/controls/Controls.jsx
RobertGemmaJr: I caught that too -it's just DOM order now but better safe than sorry

README.md
```diff
@@ -1,3 +1,35 @@
...
+## Development
+
+Clone repo and run `npm install`
+
+Run `npm start` in one terminal and `cd example && npm start` in a second terminal. The first terminal will detect changes in the library and the second will deploy the example on [http://localhost:3000](http://localhost:3000)
+
+## Install
+
+```bash
+npm install --save react-volume-viewer
```
mcmcgrath13, line 16:
```suggestion
npm install react-volume-viewer
```

example/README.md
```diff
@@ -0,0 +1,5 @@
+This example was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
+
+It is linked to the react-volume-viewer package in the parent directory for development purposes.
```
mcmcgrath13, line 3: maybe throw a screenshot on here of what it is expected to look like?

example/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "eject": "node ../node_modules/react-scripts/bin/react-scripts.js eject"
+  },
+  "dependencies": {
+    "react": "file:../node_modules/react",
+    "react-dom": "file:../node_modules/react-dom",
+    "react-scripts": "file:../node_modules/react-scripts",
+    "react-volume-viewer": "file:.."
+  },
+  "devDependencies": {
+    "@babel/plugin-syntax-object-rest-spread": "^7.8.3"
```
mcmcgrath13, line 19: why doesn't point to the root project as well?

example/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "eslintConfig": {
+    "extends": "react-app"
+  },
+  "browserslist": [
+    ">0.2%",
+    "not dead",
+    "not ie <= 11",
+    "not op_mini all"
+  ]
```
mcmcgrath13, line 29: I can't comment on binary files, but do we need all of the images in public here? What's the thinking around how much the example app should be? I was initially thinking very bare bones just renders the box and maybe some text above/below, but if the how to would be useful, maybe it's worth keeping as well? hmm

example/src/App.js
```diff
@@ -0,0 +1,41 @@
...
+        <h1>Hello, World</h1>
+      </header>
+
+      <main>
+        <VolumeViewer
+          colorMaps={{
+            Haline: haline,
+            Thermal: thermal,
+          }}
+          colorMap={haline}
```
mcmcgrath13, line 23: can you use `colorMap` without `colorMaps`? so if a app wanted to control the color map, they could and not show any options on the control bar? (or maybe not show the control bar at all, but still control the color map from the outside)

example/src/App.js
```diff
@@ -0,0 +1,41 @@
...
+          controlsVisible={true}
+          path={model}
```
mcmcgrath13, line 25: right now it looks like model is an image, not a path - should it be a path?

example/src/App.js
```diff
@@ -0,0 +1,41 @@
...
+            max: 33.71,
+            units: "°C",
+          }}
+          scale="1 -1 1"
+          rotation="-55 0 0"
+        />
+      </main>
+
+      <footer>
+        <h1>Goodbye, World!</h1>
```
mcmcgrath13, line 37: ❤️  the header/footer

example/src/App.test.js
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import App from "./App";
+
```
mcmcgrath13, line 4: can we delete most of the color maps in `example/src/assets/colormaps`?

mcmcgrath13: same with models

package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "react-volume-viewer",
+  "version": "1.0.0",
```
mcmcgrath13, line 3: I'd put our version as sub-1 right now, thoughts?

package.json
```diff
@@ -0,0 +1,70 @@
...
+    "deploy": "gh-pages -d example/build",
+    "test": "cross-env CI=1 react-scripts test --env=jsdom && run-s test:unit test:lint test:build",
+    "build": "rollup -c",
+    "start": "rollup -c -w",
+    "prepublish": "npm run build",
+    "lint": "prettier --write src example/src"
+  },
+  "dependencies": {
+    "@reach/listbox": "^0.16.2",
+    "aframe": "git+https://github.com/brown-ccv/aframe.git",
```
mcmcgrath13, line 30: this git dependency might become hairy for an independent package -  have we tried seeing if the latest aframe works?

src/Aframe/ccvLibVolumeShader.js
```diff
@@ -0,0 +1,336 @@
+/* globals  THREE */
+THREE.ShaderLib["ccvLibVolumeRenderShader"] = {
+  uniforms: {
+    //"u_mvp": { value: new THREE.Matrix4() },
+    slice: { value: 1.0 },
```
mcmcgrath13, line 5: how does this C++ injection work? Might be a question for Camilo

src/components/AframeScene.jsx
```diff
@@ -0,0 +1,90 @@
...
+import styled from "styled-components";
+import { Entity, Scene } from "aframe-react";
+
+import "aframe";
+import "aframe-event-set-component";
+import "aframe-orbit-controls";
+
+import "../Aframe/loader.js";
+import "../Aframe/buttons-check.js";
+import "../Aframe/cursor-listener";
```
mcmcgrath13, line 11: I think there were a couple other files in the `Aframe` folder not imported here - does that mean those aren't used?

src/components/AframeScene.jsx
```diff
@@ -0,0 +1,90 @@
...
+import "../Aframe/render-2d-clipplane";
+
+const StyledScene = styled(Scene)`
+  position: relative;
```
mcmcgrath13, line 15: I don't think we want to set the height/weight of the scene in any scene here, but rather set them to 100% and let the consuming application set the height of the container as they want

src/components/AframeScene.jsx
```diff
@@ -0,0 +1,90 @@
...
+      transferFunctionNodes.forEach((node) => {
+        coordinates.push(node[plane]);
+      });
+    } else console.error("Invalid Plane", plane);
+
+    return coordinates;
+  }
+
+  return (
+    <StyledScene id="volumeViewerScene" background="color: black" embedded>
```
mcmcgrath13, line 32: if one renders multiple of the volume viewers in a page, will he `id` here become a problem since it won't be unique?

src/components/App.jsx
```diff
@@ -0,0 +1,92 @@
+import React, { useEffect } from "react";
+import styled from "styled-components";
```
mcmcgrath13, line 2: the component libraries I've seen usually don't have an `App.jsx` since it's a library not an app, should this be `VolumeViewer.jsx` maybe?

src/components/App.jsx
```diff
@@ -0,0 +1,92 @@
...
+          path: path,
+          position: position,
+          rotation: rotation,
+          scale: scale,
+          slices: slices,
+          spacing: spacing,
+        }}
+      />
+
+      <div id="modelLoaded" style={{ display: "block" }}>
```
mcmcgrath13, line 73: maybe remove this div and add an issue on a loading indicator

src/components/App.jsx
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  // TODO: Add loading spinner centered on scene (could leave in AframeScene?)
+  return (
+    <Wrapper>
+      <AframeScene
+        state={state}
+        useTransferFunction={useTransferFunction}
+        model={{
+          dataRange: dataRange,
```
mcmcgrath13, line 63: I think we can just list the key here since the value is a variable of the same name e.g. `{ dataRange, ...}`

src/components/controls/ColorMapControls.jsx
```diff
@@ -0,0 +1,66 @@
...
+  ListboxArrow,
+} from "@reach/listbox";
+import "@reach/listbox/styles.css";
+
+import Title from "./SectionTitle.jsx";
+
+const StyledListboxInput = styled(ListboxInput)`
+  width: 100%;
+
+  > [data-reach-listbox-button] {
```
mcmcgrath13, line 18: is this the recommended way to reach into reach components for styling?

src/components/controls/ColorMapControls.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+    > [data-reach-listbox-arrow] {
+      margin: auto;
+    }
+  }
+`;
+
+export default function ColorMapControls({ state, setState, colorMaps }) {
+  return (
+    <div>
```
mcmcgrath13, line 32: if colorMaps is an empty object, what should happen here? An empty drop down? Return an empty div?

src/components/controls/Controls.jsx
```diff
@@ -0,0 +1,57 @@
...
+  position: absolute;
+  top: 25px;
+  left: 25px;
+  width: 300px;
+  padding: 15px;
+
+  background-color: white;
+  border-radius: 5%;
+
+  > * {
```
mcmcgrath13, line 18: Reaching into the components like this feels a bit like an antipattern (but I'm also a novice at styled components and could be dead wrong).

I also wonder if we want to hardcode the width here? or should we make it a little more reactive? Should the height be capped to the height of the volume viewer component (might already be), and if so, should this scroll?

src/test/index.test.js
```diff
@@ -0,0 +1,7 @@
+import ExampleComponent from "..";
+
```
mcmcgrath13, line 2: delete? or add tests (adding tests is probably the wiser mose)

mcmcgrath13 (commented):
This is progressing nicely!

src/components/controls/Controls.jsx
mcmcgrath13: check out https://courses.joshwcomeau.com/css-for-js/03-components/15-single-source

example/README.md
RobertGemmaJr: Good idea! I'll use a separate issue (#15) to start figuring out what should actually be in the READMEs

example/package.json
RobertGemmaJr: I think the babel dependency came from accidentally installing it in the `example` folder and not the library. It's not being used, I'll remove it.

example/src/App.js
RobertGemmaJr: I think that makes sense, yeah. I was also thinking that if `useDefaultColorMaps=false` and `colorMaps=[]` then the user has to pass in a `colorMap`.

I think we need a place to think through all these combinations, should I create an issue for it?

RobertGemmaJr: I made issue #16 we can keep tweaking things in there - I'm definitely not concrete on some of the design decisions set up so far

example/src/App.test.js
RobertGemmaJr: Didn't notice the duplicates, good idea. I think having a few different models will be good down the road so we can test "live" switching between the models. I'm still not entirely sure how the library will react to the user changing the props they send

package.json
RobertGemmaJr: Haven't tried it yet, perhaps we could make that change when we remove `aframe-react`? #5

src/Aframe/ccvLibVolumeShader.js
RobertGemmaJr: No idea, cc @kmilo9999

src/components/AframeScene.jsx
RobertGemmaJr: Amazing catch, thank you thank you thank you! Was missing `arcball-camera`. Boy did that just save many hours of headaches

src/components/AframeScene.jsx
RobertGemmaJr: Yeah I think it is, just tested it and the second volume viewer isn't displaying

src/components/App.jsx
RobertGemmaJr: The only issue is that the component in `index.js` is called `VolumeViewer` so we'd have a conflict. I could rename `App.jsx` as `Wrapper.jsx`? Or rename the component in `index.js` as `Index` (`Library`?) and just rename the export as `VolumeViewer`?

src/components/App.jsx
RobertGemmaJr: Done - issue #9. I don't like how the Spinner was rendered before, perhaps we can take another crack at it later

src/components/App.jsx
RobertGemmaJr: That worked! Is that true of all objects in JS? Very neat

src/components/controls/ColorMapControls.jsx
RobertGemmaJr: Yeah Reach talks about adding new styles [here](https://reach.tech/styling/)

src/components/controls/ColorMapControls.jsx
RobertGemmaJr: Should `colorMaps` ever be empty? Right now it's default prop is the grayscale map

src/components/controls/Controls.jsx
RobertGemmaJr: I did that so all of the spacing styles would be contained in the parent but it does make me think. Should we ditch `styled-components` and go with the `[component].module.css` files like [josh did](https://github.com/css-for-js/character-creator/tree/main/src/components/ButtonRow)?

RobertGemmaJr: cc @broarr

src/test/index.test.js
RobertGemmaJr: Do you know of any good material about testing? I've never done any JS testing before but I agree, with all the props and things going on we should setup some robust testing

package.json
RobertGemmaJr: Okay!

src/components/AframeScene.jsx
RobertGemmaJr: The user can now pass in styles directly as props - either with `className` or `styles`. The example handles the styling that way too

src/components/AframeScene.jsx
RobertGemmaJr: Actually I think it might be something else, although I'm not sure what. I'm definitely not getting any duplicate id errors

src/components/App.jsx
mcmcgrath13: Yup! It's a feature of ES6

src/components/controls/Controls.jsx
RobertGemmaJr: Actually I change my mind. I think we're going to end up building a lot of dynamic styling in which case CSS Modules would make things difficult

src/components/App.jsx
RobertGemmaJr: Something interesting I discovered - the components will re-render whenever their props change but `useState` will not be called again. By using multiple `useEffect` and the pertinent props you can get it so only changes to those props will change the internal state. Works perfectly for what we need.

PR #20, opened 2021-11-17
Refractor props

RobertGemmaJr:
Some changes to the expected/default props and a fix to the Opacity Controls. Isabel said it's best to leave the props as is, these are just the little changes I had in the works. Looking to publish the package and get RIDDC working next, then worry about prop issues as other websites  use the package.


src/components/AframeScene.jsx
```diff
@@ -12,10 +12,12 @@ import "../Aframe/cursor-listener";
 import "../Aframe/loader";
 import "../Aframe/render-2d-clipplane";
 
-function AframeScene({ state, useTransferFunction, model }) {
-  function getCoordinates(plane) {
+function AframeScene({ state, useTransferFunction }) {
+  const { model, colorMap, transferFunction, sliders } = state;
+
+  function getCoordinates(transferFunction, plane) {
```
mcmcgrath13, line 18: I think this could be simplified as a map

```js
const getCoordinates = (transferFunction, plane) => transferFunction.map((point) => point[plane])
```

I would expect transferFunction to be a function based on the name, but the usage here implies it's an array.  Might be worth thinking about variable names/types here.

src/components/App.jsx
```diff
@@ -22,31 +22,63 @@ function App(props) {
...
-    slices,
-    spacing,
+    transferFunction,
+    modelRange,
+    modelPath,
+    modelPosition,
+    modelRotation,
+    modelScale,
+    modelSlices,
+    modelSpacing,
```
mcmcgrath13, line 32: in the state it looks like you nest these all under model - should that be the same here so that passing around is easier?

src/components/App.jsx
```diff
@@ -22,31 +22,63 @@ function App(props) {
...
+      position: modelPosition,
+      rotation: modelRotation,
+      scale: modelScale,
+      slices: modelSlices,
+      spacing: modelSpacing,
+    },
     colorMap:
       colorMap && useTransferFunction ? colorMap : defaultColorMaps.Grayscale,
-    transferFunction: useTransferFunction ? initTransferFunction : null,
+    transferFunction: useTransferFunction ? transferFunction : [],
```
mcmcgrath13, line 50: could someone say 'useTransferFunction' but not supply 'transferFunction'? Maybe add a default to the object instead then just map `tranferFunction` here directly

src/components/App.jsx
```diff
@@ -22,31 +22,63 @@ function App(props) {
...
     sliders: {
       x: [sliderRange.min, sliderRange.max],
       y: [sliderRange.min, sliderRange.max],
       z: [sliderRange.min, sliderRange.max],
     },
   });
 
+  // Change model on props change
+  useEffect(() => {
```
mcmcgrath13, line 59: are both the props and state needed? or could you just pass down the props as needed? It feels a little funny to me to have an effect that's just setting state - usually effects are for side effects.

src/components/controls/Controls.jsx
```diff
@@ -17,11 +17,7 @@ function Controls(props) {
       />
 
       {state.transferFunction && (
-        <OpacityControls
-          state={state}
-          setState={setState}
-          dataRange={dataRange}
-        />
+        <OpacityControls state={state} setState={setState} />
```
mcmcgrath13, line 20: All of state really isn't needed here - consider making more targeted handlers

src/components/controls/OpacityControls.js
```diff
@@ -111,11 +111,12 @@ export default class OpacityControls extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
+    const range = nextProps.state.model.range;
```
mcmcgrath13, line 114:
```suggestion
    const { min, max, units} = nextProps.state.model.range;
```
then use those below

src/index.js
```diff
@@ -7,57 +7,92 @@ function VolumeViewer(props) {
...
 }
 
-// TODO: String validation for position, rotation, scale
-// TODO: Force user to pass min and max but not unit for dataRange?
-// TODO: Combine model into a single object? - What happens when one property changes?
-// TODO: ColorMap should be passed as a string, not the path itself - access width colorMaps[colorMap]
 VolumeViewer.propTypes = {
-  colorMap: PropTypes.string, // The current color map (import image and pass that)
+  /** The current color map (path to the image) */
+  colorMap: PropTypes.string,
```
mcmcgrath13, line 12: could these prop types be moved into what is currently called `App`, then rename that component to `VolumeViewer` and have this index.js just re-export that?  The `App` naming seems off given this isn't an app.

RobertGemmaJr: Would that make the App.jsx file too long? I just liked having the prop handling outside of the component files but they could could be consolidated

src/components/AframeScene.jsx
RobertGemmaJr: Would `transferFunctionPoints` be better? I'll add the map

src/components/App.jsx
RobertGemmaJr: It would be ideal but we lose the ability to inject default values using `PropTypes`. If they were a single object and a user passed in 
```js
<VolumeViewer
    model={
        path="summer_salt_low.png"
        slices=55
    }
/>
```

There's no way to inject the default values for `model.position`, `model.rotation`, etc. We'd essentially be forcing the user to pass in all of the model props no matter what

src/components/App.jsx
RobertGemmaJr: Yes, there's a default value for `transferFunction` that will be used if the user doesn't pass in `transferFunction` as a prop. It's at the bottom of index.js, right now the default is what the transfer function in RIDDC starts as

src/components/App.jsx
RobertGemmaJr: `useState` handles all of the state for the controls, you can't update the prop values up the component hierarchy. Having them as props is a way for the user to override the values  without physically clicking on the VolumeViewer controls. Like how with RIDDC changing from salinity to temperature models automatically changes the color map.

I'm not entirely sure why the `useEffect` is needed but I couldn't get user prop changes to update state without them. I'm thinking it's because the <App> component doesn't have to rerender, only one (or more) of it's children? Not entirely sure though.

src/components/controls/Controls.jsx
RobertGemmaJr: How would I call `setState` then? I'm calling `setState` as `setState({ ...state, transferFunction: this.transferFunctionNodes });` to not override the state values that don't change. I would have to use a separate `useState` hook for each state variable no?

src/components/controls/OpacityControls.js
RobertGemmaJr: Done (:

src/index.js
mcmcgrath13: if App.jsx is where these props are actually used, then I think it makes sense to house them there.  But again, I really don't think we should have a file named App.jsx in a component library, that file should be VolumeViewer and then just re-export it in index.js.

src/components/App.jsx
mcmcgrath13: could you use the shape check?

> // An object taking on a particular shape
  optionalObjectWithShape: PropTypes.shape({
    optionalProperty: PropTypes.string,
    requiredProperty: PropTypes.number.isRequired
  }),

src/index.js
RobertGemmaJr: Figured out a nice way of doing things... index.js just imports `VolumeViewer` (what was App.jsx) and attaches the `propTypes` and `defaultProps`

src/components/controls/OpacityControls.js
```diff
@@ -111,11 +111,12 @@ export default class OpacityControls extends Component {
...
+    const { min, max, units } = nextProps.state.model.range;
     this.dataSpace = {
-      min: nextProps.dataRange.min,
-      mid: (nextProps.dataRange.min + nextProps.dataRange.max) / 2,
-      max: nextProps.dataRange.max,
-      units: nextProps.dataRange.units,
+      min: min,
+      mid: (min + max) / 2,
+      max: max,
+      units: units,
```
mcmcgrath13, line 119: this was changed to `unit` in the props above - should that be updated here too?

src/components/VolumeViewer.jsx
```diff
@@ -0,0 +1,180 @@
...
+  useEffect(() => {
+    setState({
+      ...state,
+      transferFunction: useTransferFunction ? transferFunction : [],
+    });
+  }, [useTransferFunction, transferFunction]);
+
+  return (
+    <Wrapper className={className} style={style}>
+      <AframeScene state={state} useTransferFunction={useTransferFunction} />
```
mcmcgrath13, line 79: thinking forward, we'll want the state here to be broken down into the needed props.  I think that this might clean up some of the useEffect things as well maybe? Future work, not for now.

mcmcgrath13 (approved):
one question on units, otherwise :shipit:

src/components/controls/OpacityControls.js
RobertGemmaJr: Oh wow can't believe I missed that, the unit wasn't being displayed at all. Fixed now

src/components/VolumeViewer.jsx
RobertGemmaJr: Issue #21

Issue #15, opened 2021-11-10
Write README for library and example


RobertGemmaJr:
- `react-doxygen` should help with the library's README - autowrite props?

- Include screenshots of the example project in its README

RobertGemmaJr:
Use RIDDC `howto` images in the example's README?

`example` README should show how to use the volume viewer, `react-volume-viewer` how to develop it

Issue #13, opened 2021-11-04
Deploy to NPM


RobertGemmaJr:
Deploy library to NPM using CJ (ccv-bot@brown.edu)

PR #24, opened 2021-11-22
Npm deployment

RobertGemmaJr:
Okay lets try this again. Left the version number as 0.1.0 and included the updated README. I'll work on the changes to opacity controls in the correct branch.

- closes issue #13 
- closes issue #15

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -5,7 +5,7 @@ name: Node.js Package
 
 on:
   release:
-    types: [created]
+    types: [released]
 
 jobs:
   build:
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -32,19 +32,3 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{secrets.npm_token}}
 
-  publish-gpr:
-    needs: build
-    runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      packages: write
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
-        with:
-          node-version: 14
-          registry-url: https://npm.pkg.github.com/
-      - run: npm ci
-      - run: npm publish
-        env:
-          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -28,10 +28,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
-        with:
-          node-version: 14
       - run: npm run predeploy
       - run: npm run deploy
-        env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_AUTH_TOKEN}}
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -15,19 +15,13 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 14
+      # Publish to npm
       - run: npm ci
       - run: npm test
-
-  publish-npm:
-    needs: build
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
-        with:
-          node-version: 14
-          registry-url: https://registry.npmjs.org/
-      - run: npm ci
+      - run: npm build
       - run: npm publish
+      # Publish example on gh pages
+      - run: npm run predeploy
+      - run: npm run deploy
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_AUTH_TOKEN}}
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,4 +1,7 @@
-name: NPM Publish
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Node.js Package
 
 on:
   release:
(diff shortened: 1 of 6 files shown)


mcmcgrath13 (changes requested):
This looks closer! Add the automation so that it goes to npm on release and I think you're good to go.

RobertGemmaJr:
Added the recommended npm-publish action but I'm not sure it's setup correctly. I don't think there is a `npm ci` script for this package? Also not sure about the tokens, I created one with the NPM account as `NPM_AUTH_TOKEN` but I don't know where it's supposed to go

mcmcgrath13:
> Added the recommended npm-publish action but I'm not sure it's setup correctly. I don't think there is a `npm ci` script for this package? Also not sure about the tokens, I created one with the NPM account as `NPM_AUTH_TOKEN` but I don't know where it's supposed to go

`npm ci` is the same as `npm install` but strictly adheres to the `package-lock.json`.  The token should be added to the repo's secrets (in settings) and then you can reference it in the action as `${ { secrets.NPM_AUTH_TOKEN }}`

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,50 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
+    types: [created]
```
mcmcgrath13, line 8:
```suggestion
    types: [released]
```

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,50 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - run: npm ci
+      - run: npm test
```
mcmcgrath13, line 19: do we have tests to run? (ideally we should, but we might not yet)

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          registry-url: https://registry.npmjs.org/
+      - run: npm ci
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.npm_token}}
+
+  publish-gpr:
```
mcmcgrath13, line 35: let's skip this job, gpr is annoying to deal with

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          registry-url: https://npm.pkg.github.com/
+      - run: npm ci
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
mcmcgrath13, line 50:
```suggestion
```

.github/workflows/npm-publish.yml
mcmcgrath13: this also seems redundant with the next job, so let's delete it

mcmcgrath13 (changes requested):

.github/workflows/npm-publish.yml
RobertGemmaJr: There is an index.test.js but there aren't any real tests yet:

```js
import VolumeViewer from "..";

describe("VolumeViewer", () => {
  it("is truthy", () => {
    expect(VolumeViewer).toBeTruthy();
  });
});
```

I suppose we keep it and add in the real tests later?

.github/workflows/npm-publish.yml
```diff
@@ -30,21 +30,4 @@ jobs:
       - run: npm ci
       - run: npm publish
```
mcmcgrath13:
```suggestion
      - run: npm ci
      - run: npm test
      - run: npm build
      - run: npm publish
```

Then delete the needs on line 22 and the whole build step.  Splitting this into two jobs doesn't seem beneficial in this case

mcmcgrath13: I'm assuming we need a build step and publish doesn't handle that, but I could be wrong - you know the code best

mcmcgrath13 (changes requested):

RobertGemmaJr:
Had to adjust the Jest settings to get the tests to run properly but it should be all set now, everything passes. `package.json` includes a `predeploy` and `deploy` script which builds the example project on gh-pages. It came with the `create-react-library` script, I figure we might as well keep it?

package.json
```diff
@@ -4,37 +4,37 @@
...
-    "rc-slider": "^9.7.4"
+    "aframe-react": "^4.4.0"
   },
   "peerDependencies": {
+    "@reach/listbox": "^0.16.2",
+    "d3-scale": "^4.0.2",
+    "rc-slider": "^9.7.4",
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
     "styled-components": "^5.0.0"
```
mcmcgrath13, line 40: react and react-dom are definitely peer dependencies, but the others strike me as maybe being dependencies - @broarr what do you think?

RobertGemmaJr: I know `styled-components` recommends it being a peer dependency but I'm not sure about the other 3

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - run: npm run predeploy
+      - run: npm run deploy
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_AUTH_TOKEN}}
```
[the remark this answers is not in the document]
RobertGemmaJr: It looks like the github token `${{ secrets.GITHUB_TOKEN }}` is automatically added to the project so I just removed them.

package.json
RobertGemmaJr: Yeah I think it's best the others stay as dependencies, I'll revert that

.github/workflows/npm-publish.yml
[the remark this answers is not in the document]
RobertGemmaJr: I was able to run both `npm run predeploy` and `npm run deploy` from my terminal. I assume it's because I'm already logged in to github there but I don't know if the action will need a separate token

mcmcgrath13 (approved):
Let's give it a shot! I'm 99% sure the CI won't work as expected on the first try (because it almost never does), so be prepared to debug the first release a bit after this still.

PR #30, opened 2021-12-01
Npm deployment

RobertGemmaJr:
Temporarily changed the action to run on this branch while I try to get the CI working properly


RobertGemmaJr:
Okay everything is working correctly as of now. What do you think of combing the two jobs? It's a lot of repeated steps but suppose it might still make sense to have them be different? I left the code commented out for now before we make a decision.

Also, is there a way to condense all of these commits? Like turn them all into 1 single commit that adds the .yml file?

mcmcgrath13:
> Also, is there a way to condense all of these commits? Like turn them all into 1 single commit that adds the .yml file?

When you hit the merge button, hit the drown down arrow instead and select "squash and merge" - then this whole PR will be one commit.

.github/workflows/npm-publish.yml
```diff
@@ -1,43 +1,64 @@
...
+        run: npm run build
+      - name: Publish to NPM  
+        run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_AUTH_TOKEN}}
+      
+        # Publish to gh-pages
+      - name: Build example 
+        run: npm run predeploy
+      - name: Publish example on gh-pages
```
mcmcgrath13, line 36: this seems reasonable to me, just delete the second job

PR #31, opened 2021-12-03
Test multiple changes

RobertGemmaJr:
Fixes the issue where changing the `colorMap` and `model` props at the same time (in the example) only updated one or the other. The issue was that the `setState` calls in `useEffect` (`VolumeViewer.jsx`) weren't retrieving the most up to date state.

Also changed the default transfer function to be `[ { x: 0, y: 0 }, { x: 1, y: 1 }]` instead of `[]`. Passing an empty array makes the model disappear. `useTransferFunction=false` still uses the default transfer function for the same reason. I think the "off" should mean a perfectly linear transfer function but let me know if that's not the best UX.

A final minor change is that the color maps that the user passes down as a prop now appear at the top of the drop down in the controls, they used to be last.


src/components/controls/ClipControls.jsx
```diff
@@ -6,15 +6,15 @@ import "rc-slider/assets/index.css";
 import Title from "./SectionTitle.jsx";
 
 // TODO: Slider range should be 0-1000 with a step of 1 (changes needed in AframeScene)
-function ClipControls({ state, setState, sliderRange }) {
+function ClipControls({ state, setState, SLIDER_RANGE }) {
```
mcmcgrath13, line 9: if slider range is a const, should it be imported instead of passed around?

mcmcgrath13: or if it could be changed, maybe this prop should be `sliderRange`, but your wrapper happens to pass `SLIDER_RANGE` into it?

src/components/controls/Controls.jsx
```diff
@@ -6,7 +6,8 @@ import OpacityControls from "./OpacityControls.js";
 import ClipControls from "./ClipControls.jsx";
 
 function Controls(props) {
-  const { state, setState, sliderRange, colorMaps } = props;
+  const { state, setState, colorMaps, SLIDER_RANGE, useTransferFunction } =
+    props;
```
mcmcgrath13, line 10: in some functions you destructure in the function signature and some the next line.  I think I prefer within the signature, but mostly I prefer consistency.

mcmcgrath13 (approved):
Looks good! A couple thoughts/questions, but nothing blocking

src/components/controls/ClipControls.jsx
RobertGemmaJr: Good point - I separated all the constants into their own file. I think we may find different values to be better down the line but  they'll always be the same application wide

Issue #2, opened 2021-10-26
Refractor OpacityControls.js to functional React


RobertGemmaJr:
Copying issue over from [RIDDC](https://github.com/ridatadiscoverycenter/react-aframe-volume-renderer/issues/60).

Initial transfer function should be a prop. Default should be the same default as RIDDC. Can use React event listeners instead of pure JavaScript.

`<Controls />` should have a fixed width and the size of the canvas determined from there. Currently 250px.

Reset button should move the model back to the original position, rotation, and scale.

RobertGemmaJr:
Make sure it's not cut off when scrollbar is present

PR #33, opened 2021-12-07
Refractor opacity controls

RobertGemmaJr:
Refractors the old `OpacityControls.js` into `OpacityControls.jsx`. Updated to use React hooks, state, event listeners, etc. 

- fixes crashing error when `model.range.mid` is not provided
- reset button now resets clip and color map in addition to transfer function

closes issue #2


src/components/VolumeViewer.jsx
```diff
@@ -12,21 +13,19 @@ import {
...
+  colorMaps,
+  controlsVisible,
+  model,
+  transferFunction,
+  useDefaultColorMaps,
+  useTransferFunction,
+}) {
   const [state, setState] = useState({
-    colorMap: colorMap,
+    colorMap: colorMap ? colorMap : DEFAULT_COLOR_MAP,
```
mcmcgrath13, line 28: could use the nullish coallescer here
```suggestion
    colorMap: colorMap ?? DEFAULT_COLOR_MAP,
```

src/components/VolumeViewer.jsx
```diff
@@ -77,6 +76,7 @@ function VolumeViewer(props) {
               ? { ...colorMaps, ...DEFAULT_COLOR_MAPS }
               : colorMaps
           }
+          initColorMap={colorMap}
```
mcmcgrath13, line 79: is it a problem if this doesn't match the colorMap on line 29?

src/components/VolumeViewer.jsx
```diff
@@ -114,6 +114,7 @@ VolumeViewer.propTypes = {
     /** Minimum and maximum values of the model's dataset. Min and max values are required */
     range: PropTypes.exact({
       min: PropTypes.number.isRequired,
+      mid: PropTypes.number,
```
mcmcgrath13, line 117: why does the midpoint need to be provided? can that be calculated?

src/components/VolumeViewer.jsx
```diff
@@ -144,7 +145,9 @@ VolumeViewer.propTypes = {
 
   /**
    * Whether or not to use the libraries default color maps
-   * Default Color Maps: Grayscale, Natural, RGB
+   * Default Color Maps: grayscale, natural, rgb
+   *
```
mcmcgrath13, line 149: is this true with the `DEFAULT_COLOR_MAP` change? or is that value set to grayscale?

src/components/controls/Controls.jsx
```diff
@@ -34,7 +42,7 @@ const Wrapper = styled.div`
   margin: auto 0;
   width: 300px;
   height: fit-content;
-  max-height: calc(100% - 80px);
+  max-height: calc(100% - 50px);
```
mcmcgrath13, line 45: what is the 50px for? does it look bad if the height matches the canvas height?

src/components/controls/OpacityControls.jsx
```diff
@@ -0,0 +1,291 @@
...
+  y: [undefined, CANVAS_PADDING], // Height - padding to padding
+};
+
+// Transform transferFunction space to canvas space
+const scaleTransferFunctionToCanvasX = scaleLinear();
+const scaleTransferFunctionToCanvasY = scaleLinear();
+
+// Returns the mouse's  position relative to canvas
+function getRelativeMousePos(e) {
+  const mouse = {
```
mcmcgrath13, line 32: this isn't really the mouse position that we're returning, but rather the dragged point's position - is there a clearer way to name the variables here?

src/components/controls/OpacityControls.jsx
```diff
@@ -0,0 +1,291 @@
...
+    x: e.clientX - e.target.getBoundingClientRect().x,
+    y: e.clientY - e.target.getBoundingClientRect().y,
+  };
+
+  // Conform min/max position to padding
```
mcmcgrath13, line 37: why do we use manual padding instead of relying on css padding? are there cases where we draw into the padding region?

src/components/controls/OpacityControls.jsx
mcmcgrath13: I also wonder if there's a more direct way to get this x/y from the event

src/components/controls/OpacityControls.jsx
mcmcgrath13: this could probably all be turned into a one liner if we want to play some code golf 😉 
something along the lines of `mouse.x = Math.max(canvasRange.x[0], Math.min(canvasRange.x[1], mouse.x))`.  Could be worth a short utility function to get the bounded value given the variable name.

src/components/controls/OpacityControls.jsx
```diff
@@ -0,0 +1,291 @@
...
+
+    // Set ranges and transformations
+    canvasRange.x[1] = canvas.width - CANVAS_PADDING;
+    canvasRange.y[0] = canvas.height - CANVAS_PADDING;
+    scaleTransferFunctionToCanvasX
+      .domain(transferFunctionRange.x)
+      .range(canvasRange.x);
+    scaleTransferFunctionToCanvasY
+      .domain(transferFunctionRange.y)
+      .range(canvasRange.y);
```
mcmcgrath13, line 71: do these chained functions mutate?

src/components/controls/OpacityControls.jsx
```diff
@@ -0,0 +1,291 @@
...
+    const canvas = canvasRef.current;
+    const context = canvas.getContext("2d");
+    context.clearRect(0, 0, canvas.width, canvas.height);
+
+    // Draw rule on canvas's midpoint
+    context.beginPath();
+    context.strokeStyle = "rgba(0, 0, 0, 1)";
+    context.lineWidth = 1;
+    context.moveTo(canvas.width / 2, canvas.height);
+    context.lineTo(canvas.width / 2, canvas.height - 10);
```
mcmcgrath13, line 96: is the `- 10` here padding?

src/components/controls/OpacityControls.jsx
```diff
@@ -0,0 +1,291 @@
...
+    canvasPoints.map((point) => {
+      context.lineTo(point.x, point.y);
+    });
+    context.stroke();
+
+    // Draw points
+    canvasPoints.map((point) => {
+      context.beginPath();
+      if (pointHovering === point) context.fillStyle = "rgba(255, 255, 85, 1)";
+      else context.fillStyle = "rgba(255, 170, 0, 1)";
```
mcmcgrath13, line 112:
```suggestion
      context.fill = pointHovering === point ?  "rgba(255, 255, 85, 1)" : "rgba(255, 170, 0, 1)";
```

src/components/VolumeViewer.jsx
RobertGemmaJr: Ouu I didn't know that was a thing, neat JS really is full of neat little tricks

src/components/VolumeViewer.jsx
RobertGemmaJr: Good catch. That might have caused issues if the user didn't pass in a color map

src/components/VolumeViewer.jsx
RobertGemmaJr: It's not required but can be passed down. But now that you mention it I think I deleted where the midpoint was originally calculated. I'll make sure I add that back in

src/components/VolumeViewer.jsx
RobertGemmaJr: `DEFAULT_COLOR_MAPS` contains grayscale, natural, and rgb so yes. I could make it more explicit that they're referenced as the constant though?

mcmcgrath13: nah - I think that's an implementation detail.  I was more referring to the map (singular) since that default was introduced in this PR.

src/components/controls/Controls.jsx
RobertGemmaJr: imo it looks bad because it's just a flat white. Perhaps if we added some shadowing and what not it would still look like it was floating off the page?

src/components/controls/OpacityControls.jsx
RobertGemmaJr: I mean it is the position of the mouse, just it's position relative to the canvas's x/y coordinates... there already is a `pointDragging`. This is just a "temporary" variable used to update `pointDragging`. I get what you're saying but I'm not sure what'd work best. Just `newPosition` or something like that?

src/components/controls/OpacityControls.jsx
RobertGemmaJr: The drawing of the first and last circles get cut off without the padding because they're centered at (0, 0) and (width, height). I remember seeing some min/max stuff there before I'll have to take a second crack at it!

src/components/controls/OpacityControls.jsx
RobertGemmaJr: I'm not sure what you mean? They don't change once set but need React to render the canvas before it can grab the width and height

src/components/controls/OpacityControls.jsx
RobertGemmaJr: It's drawing a line from `canvas.height` to `canvas.height - 10` - (0,0) is in the top left of the canvas. The fact that it's the size of the padding is just a coincidence but I suppose we could set it to `canvas.height - CANVAS_PADDING` if we always wanted it to be the height of the padding?

src/components/VolumeViewer.jsx
RobertGemmaJr: I'll throw in another sentence mentioning that it'll default to grayscale if no `colorMap` is passed in

src/components/controls/OpacityControls.jsx
mcmcgrath13: I would have expected the result of these method calls to need to be saved to the variable again.  But based on usage it seems like the methods mutate the interpolator - is that the case?

RobertGemmaJr: Yeah I'm not entirely sure about the underlying architecture but it seems to be some sort of mutable function. I wrote it off [these](https://observablehq.com/@d3/d3-scalelinear) docs. The math is definitely working correctly

src/components/controls/OpacityControls.jsx
RobertGemmaJr: I changed a couple of the variable names when calling `getRelativeMousPos` to be more meaningful to what that position actually represents but left the function name the same

src/components/VolumeViewer.jsx
mcmcgrath13: If the midpoint isn't the middle, will that cause a non-linear scaling for the transfer function? If not, I don't think we should allow it as a prop at all.

src/components/controls/Controls.jsx
mcmcgrath13: I trust your style judgement - maybe just throw a comment on the css?

src/components/controls/OpacityControls.jsx
mcmcgrath13: It seems reasonable that would be the height of the padding, and if not - maybe just throw a comment that we're drawing the midpoint tick on the axis?

src/components/VolumeViewer.jsx
```diff
@@ -41,7 +41,7 @@ function VolumeViewer({
   useEffect(() => {
     setState((state) => ({
       ...state,
-      colorMap: colorMap,
+      colorMap: colorMap ?? DEFAULT_COLOR_MAP,
```
mcmcgrath13, line 44: should the calculated colorMap be its own state? or can we use `state.colorMap` instead?  This repetition makes me nervous that we'll miss something someday

src/components/VolumeViewer.jsx
RobertGemmaJr: I don't think it would really make sense to change `transferFunction` when we expose `transferFunction` itself as a prop. I hid the midpoint but I think it's better to calculate the midpoint in `VolumeViewer.jsx` instead of `OpacityControls.jsx`

src/components/controls/Controls.jsx
RobertGemmaJr: Done!

src/components/controls/OpacityControls.jsx
RobertGemmaJr: The comment is a few lines up, I made the wording a little clearer

src/components/VolumeViewer.jsx
RobertGemmaJr: `colorMap`, `model`, `sliders`, and `transferFunction` will all get their own state when we move to separate exports . I did add a couple functions for right now so there's less repeated code. For `colorMap` I was able to just add `DEFAULT_COLOR_MAP` as the default prop so there isn't any addition logic needed anymore

RobertGemmaJr:
I also made a small change here - if multiple `colorMaps` aren't present the dropdown in `ColorMapControls.jsx` becomes a static image instead of the dropdown. I got distracted and did a lot of work around the logic between `useDefaultColorMaps`, `colorMaps`, and `colorMap` but that's all in a seperate PR

src/components/VolumeViewer.jsx
mcmcgrath13: The midpoint as far as I can see is only used for drawing the tick on the axis, so I do think that should be internal to `OpacityControls.jsx`.  What's the argument for it to be in `VolumeViewer.jsx`?

src/components/VolumeViewer.jsx
```diff
@@ -24,42 +24,54 @@ function VolumeViewer({
   useDefaultColorMaps,
   useTransferFunction,
 }) {
+  function getColorMap() {
+    return colorMap;
+  }
```
mcmcgrath13, line 29: what is the use case for this function vs just using `colorMap` directly?

src/components/VolumeViewer.jsx
mcmcgrath13: For example, what if we decide we want two axis ticks instead of one down the road? Or base it on the width of the opacity viewer - I think we have the most flexibility if we keep that an implementation detail of `OpacityControls`

RobertGemmaJr: Yeah that's true. Looked a bit cleaner having it in `VolumeViewer.jsx` with all the other getter functions imo but even if we use the range to modify the transfer function directly it's still only ever used in `OpaciotyControls`.

src/components/VolumeViewer.jsx
RobertGemmaJr: The logic gets more complicated down the road. Seperate [PR](https://github.com/brown-ccv/react-volume-viewer/pull/40) but basically if the user passes in `colorMaps` but no `colorMap` it will pick the first one instead of grayscale

src/constants/constants.js
```diff
@@ -2,12 +2,16 @@ import grayscale from "../assets/grayscale.png";
...
 
+/** Default Props */
+
 const DEFAULT_COLOR_MAPS = {
-  Grayscale: grayscale,
-  Natural: natural,
-  RGB: rgb,
+  grayscale: grayscale,
+  natural: natural,
+  rgb: rgb,
```
mcmcgrath13, line 10: with the change to these to match the variable name, can use the `{grayscale, ...}` syntax

mcmcgrath13 (approved):
🙌

PR #35, opened 2021-12-08
Upgrade aframe and remove aframe-react

RobertGemmaJr:
Upgrades `aframe` to its latest versions and removes `aframe-react` dependency


RobertGemmaJr:
Merged `main` back into `upgrade-aframe` so this branch should be good to go now

RobertGemmaJr:
Also begins using `styled-components` in the example app

package.json
```diff
@@ -27,10 +27,9 @@
   },
   "dependencies": {
     "@reach/listbox": "^0.16.2",
-    "aframe": "git+https://github.com/brown-ccv/aframe.git",
+    "aframe": "^1.2.0",
```
mcmcgrath13, line 30: 🎉

mcmcgrath13: have you checked this with a headset? I think the subtle reason for the fork/pin was related to webgl versions which were used with headsets

src/Aframe/loader.js
```diff
@@ -392,12 +384,19 @@ AFRAME.registerComponent("loader", {
         data: null,
       };
 
+      // Re-inject local image with comma
```
mcmcgrath13, line 387:
```suggestion
      // Re-inject local image with semi-colon
```

src/components/AframeScene.jsx
```diff
@@ -12,31 +10,52 @@ import "../Aframe/cursor-listener";
...
-  useTransferFunction,
-}) {
+function AframeScene(props) {
+  const {
+    state: { colorMap, model, sliders, transferFunction },
+    useTransferFunction,
+  } = props;
+
+  // aframe data is passed as a string
+  function toAframeString(obj) {
```
mcmcgrath13, line 20: can this be a fat arrow function?

src/components/AframeScene.jsx
```diff
@@ -47,38 +66,38 @@ function AframeScene({
...
+      <a-entity
         id="volumeCube"
         class="clickableMesh"
-        loader={{
-          useTransferFunction: useTransferFunction,
-          colorMap: colorMap,
+        loader={toAframeString({
           alphaXDataArray: transferFunction.map((p) => p["x"]),
           alphaYDataArray: transferFunction.map((p) => p["y"]),
+          colorMap: colorMap,
```
mcmcgrath13, line 76: since this is an object, I think you could use this syntax (same with `useTransferFunction`)
```suggestion
          colorMap,
```

mcmcgrath13 (approved):
overall, looks great! A couple of the changes here, I think I've also seen in other PRs - does the merge order matter?

src/components/AframeScene.jsx
broarr: looks like that should work

broarr: Does this function need to be in this closure? Seems like you could put this outside of the component, which would make this utility function easier to test

src/components/AframeScene.jsx
```diff
@@ -12,31 +10,52 @@ import "../Aframe/cursor-listener";
...
+
+      str += propStr;
+    });
+    return str;
+  }
   return (
-    <StyledScene id="volumeViewerScene" background="color: black" embedded>
-      <Entity
+    <a-scene id="volumeViewerScene" background="color: black" embedded>
+      {/* HAND */}
```
broarr, line 36: What's `/* HAND */`?

src/components/VolumeViewer.jsx
```diff
@@ -24,15 +24,15 @@ function VolumeViewer({
   useDefaultColorMaps,
   useTransferFunction,
 }) {
-  function getColorMap() {
-    return colorMap;
-  }
-  function getModel() {
+  const getColorMap = useCallback(() => {return colorMap}, [colorMap]);
```
broarr, line 27: out of curiosity, why did you choose to use a callback here? Is this to prevent recalculating the color map or something? I'm also curious about `getModel` and `getTransferFunction`

src/components/VolumeViewer.jsx
```diff
@@ -24,15 +24,15 @@ function VolumeViewer({
...
-  function getTransferFunction() {
+  }, [model]);
+
+  const getTransferFunction = useCallback(() => {
     return useTransferFunction ? transferFunction : DEFAULT_TRANSFER_FUNCTION;
-  }
+  }, [useTransferFunction, transferFunction])
 
   const [state, setState] = useState({
     colorMap: getColorMap(),
```
broarr, line 38: I think I'm a bit confused. If you already have the transfer function/color map/model as props, why do you need to store it in state wrapped in a callback? What's the usecase? Are you trying to prevent rerendering the volume viewer if the props change?

broarr (commented):
I left some questions on the Volume Viewer component. I don't fully understand what you're trying to do there. Hit me up if a synchronous convo would be easier!

src/components/AframeScene.jsx
RobertGemmaJr: Good point I'll pull it out of the component. What's the benefit of an arrow function over the normal definition?

mcmcgrath13: It creates it's own scope for `this`, so it's unambiguous what the scope is. It's also shorter 😄.  Generally, unless you need a `function`, I always use fat arrow functions.

src/components/AframeScene.jsx
RobertGemmaJr: It's a physical hand controller in VR, at least as far as I'm aware. This gets into why I'm not sure if we need aframe or if Three.js handles things fine...

src/components/VolumeViewer.jsx
RobertGemmaJr: ESLint wants the functions as dependencies in the `useEffect` hook (line 56, 64, etc) but doing so would cause the `useEffect` to constantly re-render, using the callback prevents this

RobertGemmaJr: Actually now that you mention that, I'm pretty sure I could ditch the `useEffect`s and handle everything in the callback functions

src/components/VolumeViewer.jsx
RobertGemmaJr: I want a prop change to override a state change. The values can be changed internally (in `Controls.jsx`) which  is why I'm using state - a prop change should also update that state

RobertGemmaJr:
> overall, looks great! A couple of the changes here, I think I've also seen in other PRs - does the merge order matter?

Merge order is okay for these PRs. I haven't fully setup the branches in which order does matter yet

src/components/AframeScene.jsx
RobertGemmaJr: That worked (:

src/components/VolumeViewer.jsx
RobertGemmaJr: I'll handle that in #43 though since that PR is what's actually working with those variables

src/components/AframeScene.jsx
RobertGemmaJr: Make's sense! I updated it

PR #40, opened 2021-12-15
Color maps logic

RobertGemmaJr:
**colorMap**
- The `colorMap` prop if given
- The first property in `colorMaps` if `colorMap` is not given but `colorMaps` is
- `DEFAULT_COLOR_MAP` (grayscale) if neither `colorMap` nor `colorMaps` are given.

**colorMaps**
I added some logic to make sure `colorMap` is added to `colorMaps` if the user doesn't do so. And with that knowledge:

- The `colorMaps` prop + `DEFAULT_COLOR_MAPS` if `colorMaps` is given and `useDefaultColorMaps` is true
- Just `colorMaps` if `colorMaps` is given and `useDefaultColorMaps` is false. 
- Just `DEFAULT_COLOR_MAPS` if `colorMaps` is not given and `useDefaultColorMaps` is true.
- Just `DEFAULT_COLOR_MAP` (`{default: grayscale}`) if `colorMaps` is not given and `useDefaultColorMaps` is false.

PR #43, opened 2022-01-06
Sep state vars

RobertGemmaJr:
Separates `colorMaps`, `model`, `transferFunction`, etc. into their own state variables. This will make separating the app into different exports much easier but that's left for another PR


RobertGemmaJr:
The reset button is moved up into `Controls.jsx` and now remounts the entire component when clicked. This resets the model's position in the scene

src/components/controls/Controls.jsx
```diff
@@ -6,29 +6,35 @@ import OpacityControls from "./OpacityControls.jsx";
...
         <OpacityControls
-          state={state}
-          setState={setState}
-          initColorMap={initColorMap}
+          range={model.range}
+          initTransferFunction={initTransferFunction}
+          setTransferFunction={setTransferFunction}
         />
       )}
+      <Button onClick={reset}> Reset </Button>
```
mcmcgrath13, line 35: does reset reset everything? or just the camera angle?

mcmcgrath13 (approved):
🙌

src/components/VolumeViewer.jsx
```diff
@@ -16,74 +16,94 @@ import AframeScene from "./AframeScene.jsx";
...
-      z: [SLIDER_RANGE.min, SLIDER_RANGE.max],
-    },
-    transferFunction: getTransferFunction(),
-  });
-
-  // Update colorMap on prop change
+  // Changing a components key will remount the entire thing
+  // Because the model's position is handled internally by aframe we
+  // need to remount it to reset its position
+  const [remountKey, setRemountKey] = useState(Math.random());
```
broarr, line 30: That's an interesting way of forcing a rerender!

src/components/VolumeViewer.jsx
```diff
@@ -16,74 +16,94 @@ import AframeScene from "./AframeScene.jsx";
...
+
+  // Control colorMap in state and update on prop change
+  const getColorMap = useCallback(() => {
+    return colorMapProp ?? DEFAULT_COLOR_MAP;
+  }, [colorMapProp]);
+  const [colorMap, setColorMap] = useState(getColorMap());
```
broarr, line 36: You did this in another PR. I'm not sure I understand why you're creating a color map callback and then storing that in state. Couldn't you store `const [colorMap, setColorMap] = useState(colorMapProp ?? DEFAULT_COLOR_MAP);`?

src/components/VolumeViewer.jsx
```diff
@@ -16,74 +16,94 @@ import AframeScene from "./AframeScene.jsx";
...
   useEffect(() => {
-    setState((state) => ({
-      ...state,
-      colorMap: getColorMap(),
-    }));
-  }, [colorMap]);
-
-  // Update model on prop change
+    setColorMap(getColorMap());
```
broarr, line 38: Can this line be `setColorMap(colorMapProp ?? DEFAULT_COLOR_MAP)`? Do you have the callback function to DRY up this code?

broarr (commented):
Some questions about react stuff, but this looks good otherwise

src/components/controls/Controls.jsx
RobertGemmaJr: Yup! The state reset is defined in `VolumeViewer.jsx`

src/components/VolumeViewer.jsx
RobertGemmaJr: Stack overflow seems to agree it's the easiest way of doing so

src/components/VolumeViewer.jsx
RobertGemmaJr: The logic ends up much more complicated down the line (plenty enough to warrant a separate function imo), it's really just skeleton code at this point

RobertGemmaJr: #40

src/components/VolumeViewer.jsx
RobertGemmaJr: Yeah same as above - when things get more complicated it's a lot of logic to repeat

RobertGemmaJr:
RE the `useEffect` and `useCallback` it still seems like this is the best way to handle it. If I do `const [model, setModel] = useState({ ...DEFAULT_MODEL, ...modelProp });` for example, it won't re-fire when `modelProp` changes. It needs to be a `setModel` and in order to do that without constantly forcing a re-render it needs to be inside a `useEffect`.

The only simplification I could do is remove the function (`getModel`) from the `useEffect` dependency array. This would let me eliminate the `usecallback` step (making it a basic arrow function) but ESLint shows me a warning when the function is missing from the dependency array.

I'm learning on the fly with all this so I very well might not be looking at the best solution but I don't know any way to change the props from another component.

mcmcgrath13:
`useEffect` with the correct dependencies sounds pretty idiomatic to me.  You can put a comment on the line about the dependencies to tell eslint to ignore that error.  eslint is not all knowing 😉

RobertGemmaJr:
Actually I think I found a better solution. Having the getter function inside the closure meant React was re-creating the function definition for each render. Pulling them outside of the closure means there's a single function definition, I just need to pass the prop in as a parameter since it's no longer in the same scope. Since the function will never change it's not actually a dependency for `useEffect` and, even better, the function will get called whenever the prop changes so I can use that return to update state. It's a similar solution as `toAframeString` in the other branch and is way cleaner.

I'm not sure what to call the return variable though. They're prepended with `init` right now but I'm not sure that's the clearest name

mcmcgrath13 (approved):
🥇

PR #44, opened 2022-01-06
Fixes Volume Viewers default size

RobertGemmaJr:
The VolumeViewer will now fill 100% of the width and height of it's container by default.

Makes a good bit of changes in the example App

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -90,7 +90,7 @@ function App() {
 
       <Main>
         {Buttons}
-        <Wrapper>{VV}</Wrapper>
+        {VV}
         <hr />
       </Main>
 
@@ -101,14 +101,8 @@ function App() {
   );
 }
 
-const Wrapper = styled.div`
-  height: 250px;
-  width: 250px;
-  margin: auto;
-`;
-
 const StyledVolumeViewer = styled(VolumeViewer)`
-  /* height: 75vh; */
+  height: 75vh;
 `;
 
 const Main = styled.main`
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -1,7 +1,6 @@
 import React, { useState } from "react";
-
+import styled from "styled-components";
 import { VolumeViewer } from "react-volume-viewer";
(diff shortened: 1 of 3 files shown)


example/src/App.js
```diff
@@ -21,74 +20,99 @@ export default function App() {
...
 }
+
+const Wrapper = styled.div`
+  height: 250px;
+  width: 250px;
+  margin: auto;
+`;
+
+const StyledVolumeViewer = styled(VolumeViewer)`
+  /* height: 75vh; */
```
mcmcgrath13, line 111: why commented out?

src/components/VolumeViewer.jsx
```diff
@@ -93,6 +93,8 @@ function VolumeViewer({
 const Wrapper = styled.div`
   position: relative;
   isolation: isolate;
+  width: 100%;
```
mcmcgrath13, line 96: how does this compare to the previous behavior? doesn't a div already default to this behavior? (I'm pretty sure it does on width, less sure on height)

example/src/App.js
RobertGemmaJr: Was swapping between defining the height in the wrapper and directly (what's commented out). I'll delete it though because the issue is definitely resolved

src/components/VolumeViewer.jsx
RobertGemmaJr: Width defaults to 100% but the height is determined by the children, defaulting to 0. `aframe` sets height and width to 100% internally. 

The height was being set externally, `VolumeViewer.jsx` would default to a height of 0 and aframe would set itself to 100% of 0. We're still telling the end user to set their own height but now `VolumeViewer` becomes 100% of that height.

I added the width in even though it's the default behavior just for code readability. I feel like it makes the width/height a bit more explicit but I could delete if you think it's best to just let HTML/CSS handle it

mcmcgrath13 (approved):
🚢

broarr (approved):

Issue #46, opened 2022-01-19
Add Channel support


kmilo9999:
The basic functionality of reading volumes from a single channel was removed.  Give support to this feature using the props send to the volume viewer.

PR #48, opened 2022-01-21
Fix channels support

kmilo9999:
Solves #46
- Added changes in Aframe component and shader files to support switch between channels.
- Added optional value to increase pixel intensity.
- Fixed coded that was calling updateTransferTexture twice when a single prop was being updated


mcmcgrath13 (commented):
do the channel and intensity props need to be added to the proptypes declaration?

kmilo9999:
Prototype of the shader lib component or the Aframe component?

src/components/VolumeViewer.jsx
```diff
@@ -40,6 +40,8 @@ function VolumeViewer({
   transferFunction: transferFunctionProp,
   useDefaultColorMaps,
   useTransferFunction,
+  channel,
+  intensity,
```
RobertGemmaJr, line 44: Mary's right, these will need to be added to the Proptypes at the bottom of this file.

src/Aframe/ccvLibVolumeShader.js
```diff
@@ -272,20 +274,26 @@ THREE.ShaderLib["ccvLibVolumeRenderShader"] = {
...
-    //'discard;',
     "}",
 
     "}",
-
+    // artifitially increasing pixel intensity
+    "smple.rgb = smple.rgb*intensity;",
+    //"smple.a = smple.a*0.9;",
     //'smple.a = max(smple.r, max(smple.g,smple.b)) ; ',
     //'smple.a = 0.1*smple.a;',
```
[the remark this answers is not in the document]
kmilo9999: Done!

src/components/VolumeViewer.jsx
RobertGemmaJr: Also, `Proptypes.defaultValues`

RobertGemmaJr (changes requested):
What exactly are `channel` and `intensity` being applied to? Are they features of the model or are they features of the color map? Or are they entirely separate?

For instance is it possible to have a color map and set channel/intensity or if you set channel/intensity do you give up the ability to use colorMap? We should make sure to do validation for all this stuff. Can be done in a different branch I'm just trying to understand what channel and intensity actually effect.

src/components/VolumeViewer.jsx
kmilo9999: Good questions!

- Channel and intensity are properties of the model.  So, I assume the properties should go in the Model object of the Volume viewer component, correct ? If that's so, I can do the change.

- Intensity  is an optional value. It  can be applied to both colored mapped model and only channels, and It is only applying an extra push to the color intensity when it needs to. I had to do it for the spheroids due the data is very dim.

- We dont need to give up the color mapping for channels. I did test channel + color mapping and it seems to work. What the shader does in that case is using the maximum between the 3 channels in that pixel and looking up what color represents that coordinate in the color mapping.

RobertGemmaJr: Yeah I would move them under `model` in that case. The proptypes are at the bottom of `VolumeViewer.jsx`:

```js
  /** The model to be displayed and it's related information */
  model: PropTypes.shape({
    /** Path to the model REQUIRED */
    path: PropTypes.string.isRequired,
    /** Position of the model in the scene */
    position: PropTypes.string,
    /** Minimum and maximum values of the model's dataset. Min and max values are required */
    range: PropTypes.shape({
      min: PropTypes.number.isRequired,
      max: PropTypes.number.isRequired,
      unit: PropTypes.string,
    }),
    /** Position of the model in the scene */
    rotation: PropTypes.string,
    /** Scale of the model in the scene */
    scale: PropTypes.string,
    /** Number of slices used to generate the model */
    slices: PropTypes.number,
    /** Spacing between the slices of the model */
    spacing: PropTypes.exact({
      x: PropTypes.number,
      y: PropTypes.number,
      z: PropTypes.number,
    }),
  }),
```
and the default values are in `constants.js`:
```js
const DEFAULT_MODEL = {
  position: "0 0 0",
  range: { min: 0, max: 1, unit: "" },
  rotation: "0 0 0",
  scale: "1 1 1",
  slices: 55,
  spacing: { x: 2, y: 2, z: 1 },
};
```

The validation should work correctly once you specify the type and default values

kmilo9999: Done!!

RobertGemmaJr (approved):
Looks great! Run an `npm run lint` before you merge and everything's good to go

Also update the package version number. Say `0.2.2` (or `0.3.0` @mcmcgrath13?). Once it's merged with main we can make a new release and we'll be able to update the library in the other repos

mcmcgrath13:
> Also update the package version number. Say `0.2.2` (or `0.3.0` @mcmcgrath13?). Once it's merged with main we can make a new release and we'll be able to update the library in the other repos

This is a new feature, so I'd say `0.3.0`

mcmcgrath13:
I don't see the package number bumped in package.json, otherwise I think it's good to go.

mcmcgrath13 (approved):

PR #52, opened 2022-01-26
Fix styling on Controls menu

RobertGemmaJr:
- Removes `border-radius`
- Aligns menu to top-right of the `AframeScene`
- Sets top/right/bottom spacing to `8px`
- Adjusts margin between sections to `16px`
- Adjusts margin under headings to `8px`
- Displays `colorMap.name` in dropdown controls
- Minor version bump to 0.3.1


mcmcgrath13:
@RobertGemmaJr could you include a screenshot of what this looks like? (and maybe at a couple different viewport sizes?) Thanks!

mcmcgrath13:
Is the recommendation for how a user of this package would style things out of scope of this PR?

RobertGemmaJr:
**Typical look:**
image

**Scrolling:**
image

**Extra Room**
image

The width is still hard coded but I plans for mobile will need a lot of thought in a separate PR

RobertGemmaJr:
> Is the recommendation for how a user of this package would style things out of scope of this PR?

I think so. I think it's best to follow [Reach UI's](https://reach.tech/styling) model for developer styling in which case we'll have to expose the class names in a better way. These are just the changes we talked about in the last meeting. 

I added #8 a while go but it's been on the backburner ever since. Perhaps we start planning there?

mcmcgrath13:
That looks better to me.  I do think adding the styles customization recommendations is something that should get prioritized (and shouldn't be that hard I think)

mcmcgrath13 (approved):
🟦

PR #55, opened 2022-01-27
Clean aframe

RobertGemmaJr:
- Cleans up the files in the `aframe/` folder
- Removes 2 aframe dependencies that weren't being used. 
- The variables inside `model.js` (was `loader.js`) follow the same schema as the react variables within the component, makes the handoff much cleaner
- Reorganized the functions/logic in multiple files
- Adds code to remove event listeners when the entity is removed
- Renamed some variables/functions
- Lots of other changes I'm probably forgetting

I don't think this is all the cleanup/consolidation we could/should do but I keep coming back to this place as a baseline cleanup. The rest of the stuff I have in mind involves combining files/entities, it's a lot harder to ensure all the functionality remains the same with those changes. Plus I want to make sure the bulk of the cleanup is up to date before some higher priority tasks take over


src/Aframe/cursor-listener.js
```diff
@@ -1,42 +1,34 @@
...
 
-  onMousedown: function (evt) {
+  onMouseDown: function (e) {
     this.dragging = true;
   },
 
-  onMouseMove: function (evt) {
-    if (this.dragging === true) {
-      console.log("dragging at: ", evt.detail.intersection.point);
+  onMouseMove: function (e) {
```
mcmcgrath13, line 25: is this cursor listener actually used?

src/Aframe/model.js
```diff
@@ -0,0 +1,381 @@
...
+    img.onload = () => {
+      ctx.drawImage(img, 0, 0);
+      const colorData = ctx.getImageData(0, 0, img.width, 1).data;
+      const colorTransfer = new Uint8Array(3 * 256);
+
+      // Extract RGB values from colorMap (ignore alpha)
+      for (let i = 0; i < 256; i++) {
+        colorTransfer[i * 3 + 0] = colorData[i * 4 + 0];
+        colorTransfer[i * 3 + 1] = colorData[i * 4 + 1];
+        colorTransfer[i * 3 + 2] = colorData[i * 4 + 2];
```
mcmcgrath13, line 256: make this another loop over i = 0..2?

src/Aframe/model.js
```diff
@@ -0,0 +1,381 @@
...
+  },
+
+  updateChannel: function () {
+    const mesh = this.getMesh();
+    if (mesh) {
+      const material = mesh.material;
+      material.uniforms.channel.value = this.data.channel;
+      material.uniforms.useLut.value = this.data.useTransferFunction;
+      material.needsUpdate = true;
+    } else console.log("MODEL NOT LOADED YET");
```
mcmcgrath13, line 328: is this console log still needed?

mcmcgrath13: same with the one about 10 lines up

src/components/AframeScene.jsx
```diff
@@ -75,18 +76,17 @@ function AframeScene({
...
-          alphaYDataArray: transferFunction.map((p) => p["y"]),
-          colorMap: colorMap.path,
-          path: model.path,
-          slices: model.slices,
-          x_spacing: model.spacing.x,
-          y_spacing: model.spacing.y,
-          z_spacing: model.spacing.z,
+        model={toAframeString({
+          // colorMap: colorMap.path,
+          colorMap: JSON.stringify(colorMap),
```
mcmcgrath13, line 81: why the stringifying?

src/Aframe/cursor-listener.js
RobertGemmaJr: It's "used" but not actually changing anything. I was waiting merge this into main to start deleting files but this one is pretty obviously not changing the functionality at all. I'll delete

src/Aframe/model.js
RobertGemmaJr: Sure! There's a similar one in `updateTransferTexture` too

src/Aframe/model.js
RobertGemmaJr: I added these as some debugging for what 's coming. I suppose I could delete them now that the change is out of scope for the PR (#56)? All those functions should only be called after the model's mesh has been created

src/components/AframeScene.jsx
RobertGemmaJr: Aframe components take all their data as a string and the schema only parses relatively simple data types. Not sure why but we're kind of stuck with it. This is the way to parse the more complicated data structures we have into aframe

https://aframe.io/docs/1.2.0/core/component.html#property-types

src/Aframe/model.js
RobertGemmaJr: I deleted them and made a note where they were used in the issue

mcmcgrath13 (approved):
🧹

PR #59, opened 2022-02-24
Consolidate props

RobertGemmaJr:
- Move `colorMap` prop into `model`
- Move `transferFunction` into `model`
- Convert `colorMap` to an object internally

I tried making `colorMap` a `Map` internally but I need to be able to loop over it and the map type only has a `forEach` loop. I could try to work around it but it would move the JSX code outside of the `return` statement in `ColorMapControls.jsx`. Not sure what's a better plan?


README.md
```diff
@@ -56,44 +43,59 @@ VolumeViewer.propTypes = {
 
   /** The model to be displayed and it's related information */
   model: PropTypes.shape({
+    /** Common name of the color map applied by the transfer function. REQUIRED */
+    colorMap: PropTypes.string.isRequired,
+
+    /** Channel to load data from (R:1, G:2, B:3)*/
+    channel: PropTypes.number,
```
mcmcgrath13, line 50: I thought we always loaded from channel 1?

README.md
```diff
@@ -144,9 +146,10 @@ function App() {
   return (
     <StyledVolumeViewer
       colorMaps={[haline, thermal, ColorMaps.grayscale]}
-      colorMap={haline}
+      colorMap={haline.name}
       controlsVisible={controlsVisible}
       model={{
+        colorMap={haline},
```
mcmcgrath13, line 152:
```suggestion
      controlsVisible={controlsVisible}
      model={{
        colorMap={haline.name},
```

Is this correct?

src/components/AframeScene/AframeScene.jsx
```diff
@@ -75,15 +69,15 @@ function AframeScene({
         id="volumeCube"
         class="clickableMesh"
         model={toAframeString({
-          colorMap: JSON.stringify(colorMap),
-          sliders: JSON.stringify(sliders),
-          transferFunction: JSON.stringify(transferFunction),
-          useTransferFunction,
           channel: model.channel,
+          colorMap: JSON.stringify(colorMaps[model.colorMap]),
```
mcmcgrath13, line 73: it seems like `colorMap` is now being used somewhat interchangably to refer to the name, the whole object or the path ( in the aframe model above).  It might be worth using more verbose names to be clear about what is what.

src/components/VolumeViewer/VolumeViewer.jsx
```diff
@@ -12,84 +11,68 @@ import {
...
+    [modelProp]
+  );
+  const colorMaps = useMemo(
+    () =>
+      // Convert colorMaps to an object
+      colorMapsProp.reduce(
+        (obj, colorMap) =>
+          Object.assign(obj, { [colorMap.name]: colorMap.path }),
+        {}
+      ),
```
mcmcgrath13, line 44: is this just deep copying?

README.md
RobertGemmaJr: I'm pretty sure this is the setup for the microscopy data? CC @kmilo9999

README.md
RobertGemmaJr: Yeah, `colorMaps` is the string corresponding to the name in `colorMaps`

src/components/AframeScene/AframeScene.jsx
RobertGemmaJr: Changed the name to `colorMapSrc` in Aframe

src/components/VolumeViewer/VolumeViewer.jsx
RobertGemmaJr: Yeah. I couldn't get the `forEach` loop to return any `jsx` when using the true JS `map` type so I decided to go with just an object. I'm not entirely sure what the best way to handle it is

README.md
mcmcgrath13: that's changing/changed to 1 8-bit png per channel so the model to file ratio should be 1:1 and there will only be one channel at this time

README.md
mcmcgrath13: wait - is my suggestion correct or is the readme correct? the readme doesn't match up with my understanding of the implementation

src/components/AframeScene/AframeScene.jsx
```diff
@@ -75,15 +69,15 @@ function AframeScene({
         id="volumeCube"
         class="clickableMesh"
         model={toAframeString({
-          colorMap: JSON.stringify(colorMap),
-          sliders: JSON.stringify(sliders),
-          transferFunction: JSON.stringify(transferFunction),
-          useTransferFunction,
           channel: model.channel,
+          colorMapSrc: colorMaps[model.colorMap],
```
mcmcgrath13, line 73: it looks like the usage of `colorMapSrc` is expecting a png encoded string, but wouldn't this be a path/name pair?

src/components/VolumeViewer/VolumeViewer.jsx
```diff
@@ -12,84 +11,73 @@ import {
...
-  useTransferFunction,
 }) {
-  // Get initial values based on prop input. These will update on prop change
-  const initColorMap = getColorMap(colorMapProp, colorMapsProp);
-  const colorMaps = getColorMaps(initColorMap, colorMapsProp);
-  const initTransferFunction = getTransferFunction(
-    useTransferFunction,
-    transferFunctionProp
+  // Get initial values based on prop input. These will update on prop change.
+  const initModel = useMemo(
```
mcmcgrath13, line 25: do we want to `useMemo` on functions here, or make the whole component memoized?

src/components/VolumeViewer/VolumeViewer.jsx
mcmcgrath13: so is colorMaps not an object to start? I had assumed it was from the propTypes.  Could you update the comment to either clarify what we are converting from or that this is just deep copying and why?

README.md
RobertGemmaJr: Okay so maybe we don't need the channel data at all? Just want to make sure before I delete it

mcmcgrath13: I'm pretty sure what @mirestrepo @kmilo9999 and I had discussed was that this project would expect single channel pngs only (as a multi channel png is a pre-colored scene and (logically, though not technically) incompatible with a transfer function. That might be a different use case that comes up someday, but I don't think we're looking to support it right now.

src/components/AframeScene/AframeScene.jsx
RobertGemmaJr: Okay so this is a bit of an oddity that I'm not sure how to best work with. The color maps that we export for the user to use (Grayscale, Natural, RGB) become that png encoded string. But, the user can pass a path to a png and that will work too.

Do we expect the user to import their images first? How would we validate that? Or is it okay as it is right now?

image

src/components/VolumeViewer/VolumeViewer.jsx
RobertGemmaJr: What would the benefit be? `VolumeViewer.jsx` is really just prop validation, state management, and a wrapper component so it may be beneficial to wrap the whole thing.

mcmcgrath13: that allows a more global comparison of prop changes (could do a lodash deepequal instead) instead of doing it one by one

RobertGemmaJr: Yeah did a little research and I definitely think it will help. I think `AframeScene.jsx` should be memoized as well. Added as #60 and will implement in a seperate pr

src/components/AframeScene/AframeScene.jsx
RobertGemmaJr: I adjusted the function a little and added better comments to the `toAframeString` function that makes the functionality clearer

src/components/VolumeViewer/VolumeViewer.jsx
RobertGemmaJr: Okay so `colorMaps` is an array of `colorMap` objects:

```js
/**
 * Object containing the name and path to a color map image
 *  name: Common name of the color map
 *  path: Path to the color map source image
 */
const COLOR_MAP = PropTypes.exact({
  name: PropTypes.string,
  path: PropTypes.string,
});

  /**
   * Array of color maps available in the controls.
   *  name: Common name of the color map
   *  path: Path to the color map src
   */
  colorMaps: PropTypes.arrayOf(COLOR_MAP),
```

README.md
```diff
@@ -25,75 +25,82 @@ The only required props are the model's path and it's minimum and maximum data p
...
+ *  name: Common name of the color map
+ *  path: Path to the color map source image
+ */
+const COLOR_MAP = PropTypes.exact({
+  name: PropTypes.string,
+  path: PropTypes.string,
+});
 
+VolumeViewer.propTypes = {
   /**
```
mcmcgrath13, line 39: I'd add to the description here that the `colorMap` must be present in `colorMaps`

README.md
```diff
@@ -25,75 +25,82 @@ The only required props are the model's path and it's minimum and maximum data p
...
    *  name: Common name of the color map
    *  path: Path to the color map src
    */
-  colorMaps: PropTypes.arrayOf(
-    PropTypes.exact({
-      name: PropTypes.string,
-      path: PropTypes.string,
-    })
-  ),
+  colorMaps: PropTypes.arrayOf(COLOR_MAP),
```
mcmcgrath13, line 44: is moving colorMaps into model going to be covered by a future PR?

src/components/ColorMapControls/ColorMapControls.jsx
```diff
@@ -12,20 +12,25 @@ import "@reach/listbox/styles.css";
 
 import Section from "../Section";
 
-function ColorMapControls({ colorMaps, colorMap, setColorMap }) {
+function ColorMapControls({ colorMaps, model, setModel }) {
   return (
     <Section title="Color Map">
-      {colorMaps.length > 1 ? (
+      {Object.keys(colorMaps).length > 1 ? (
```
mcmcgrath13, line 18: this can change back to just a length comparison without keys I think?

mcmcgrath13 (approved):
couple tiny nits, but overall this looks great!

README.md
RobertGemmaJr: I don't know that we had decided that? If the dropdown was going to be all the possible colorMaps for any model or every model would have it's own colorMaps combination? The change will be easy enough so I'll do it in another PR

README.md
RobertGemmaJr: Done!

src/components/ColorMapControls/ColorMapControls.jsx
RobertGemmaJr: Good catch

README.md
mcmcgrath13: I thought we had (cc @mirestrepo @kmilo9999 ) as it might be that for a set of models we want to enforce a color map  for each model and not have the drop down, which if we have global color maps isn't possible.

Issue #57, opened 2022-02-09
Ability to take multiple models, colorMaps, and transferFunctions


RobertGemmaJr:
Microscopy dataset will overlay multiple `<a-entity>` models on top of each other, each with their own `colorMap` and `transferFunction`

The menu will need some sort of drop down to select each `model` individually and change it's `colorMap` and `transferFunction` from there

RobertGemmaJr:
@mcmcgrath13 @kmilo9999 Should we force the user to pass in an array of models and transferFunctions? Or use internal checks to see if they only passed one in?

- `model` -> `models`
- `transferFunction` -> `transferFunctions`

mcmcgrath13:
hmmmmmmmmmmm

mcmcgrath13:
I can never remember whether ctrl-enter is new line or submit... 🤦 

I'm leaning models and force it to be an array.  It makes for more consistent types/documentation.  *But*, I also imagine that's the less common use case, so maybe we shouldn't make it the default.  What's your impression on what will be used more frequently?

Issue #51, opened 2022-01-24
Seperate component for model a-entity


RobertGemmaJr:
Should be a separate react component so multiple can be added to the scene. Then `model` prop becomes `models`, an array of the `model`s

RobertGemmaJr:
Using `.map()` and aframe's own `a-entity` function accomplishes the same result. This way there's no react components within the `a-scene`

PR #62, opened 2022-03-03
Feat multiple models

RobertGemmaJr:
- Changes `model` prop to `models` array
- Each model is controlled individually with tabs in the control's menu
- Adds `name`, `description`, and `enabled` props to each model.
- Model is only displayed if `enabled === true`. No way to change in the UI controls yet.
- Each aframe component has a unique id

I do know of a couple issues though:
- Overlaying models doesn't work. We use the alpha channel to apply the color map so I'm not sure how to handle the transparency? The models look good side by side but the last model will mask all the others if they're in the same position
- Changing the `controlsVisible` prop cause `initModels` to recalculate. The `useMemo` isn't doing anything, I gather I need to do a deep comparison? It doesn't look like `React.useMemo` can taker a custom comparison function though
- Lots of unnecessary re-rendering going on. `AframeScene` re-renders 3 times when opening the controls and basically anytime you interact with it. 

I think the re-rendering issues could be pushed to a different PR but I think the `initModels` should be fixed

- closes issue #47 
- closes issue #51 
- closes issue #57


RobertGemmaJr:
Should `sliders` be unique to each model as well?

RobertGemmaJr:
Okay I think I've got a better understand of what's going on here. Any changes to the `VolumeViewer` props (from the example app) will cause the entire component to re-render. The `useMemo` that builds `initModels` runs again because it only performs a shallow comparison and `modelsProp` is an array. That causes `initModels` to change, which fires the `useEffect`, which then calls `setState(initModels)`. The `setState` call returns `models` to all of the prop values. 

- Changing `controlsVisible` (or any other prop that isn't `models`) shouldn't cause the `initModels` memo to run. Need to do a deep comparison to tell if the `models` prop actually changed
- Making a change to `models` should only make the specific change from the previous models prop in state. The change might only be 1 property of 1 specific model inside the `models` array. Need to determine the difference between `modelsProp` and the previous `modelsProp` and apply that change to `models`. Everything else about `models` should be the same after the `setModels` call. 

Looks like `useRef` is an easy way to keep track of previous values so I'll implement that. I might handle everything inside the `useEffect` because `useMemo` can't take a custom comparison function.

RobertGemmaJr:
Converting to a draft while we figure out the other issues. Those will be merged into this branch as their own PR before bringing everything into main. I will work on the blending of the models and then focus on the external state.

## Blending Models

There will only be one `<a-entity />` in the scene (`AframeScene.jsx`)
- Remove `models.map` call but keep `<a-entity />` parent component (tracks position/rotation/scale)
- `position`, `rotation`, and `scale` will be individual props of the `VolumeViewer` component. Have to be the same for all of the `png`s

`model.js` takes the full `models` array. Should rename file to `models.js`. 
- Will have to figure out working with color maps that are png encoded strings
- Aframe schema validation? Most validation will be done in the library itself so I don't think there needs to check that much additional stuff

Control flow of `models.js` will look super different. On `init` and `update` (`oldData.models != this.data.models`):
- Create the texture for each individual model (including `colorMap`, `transferFunction`, and `clipping`)
- `updateTransferTexture` should blend all of the model's textures into a single texture (webGL equivalent of `glBlendFunc`) and apply that texture to the aframe model
- The creation of the `material`, `unit cube`, and `mesh` is currently in  `loadModel`. `loadModel` should only call `THREE.TextureLoader` (and save in an array). That code gets moved to the end of `updateTransferTexture`

##  Managing State
- The controls should change some global state - not it's own internal thing
- User should load config file into state and pass that state into the library - `const [models, setModels] = useState([array loaded from config]);`. Both `models` and `setModels` need to be passed into the library
- The `setModels` calls currently being used can stay the same - just use prop instead of internal state


## Misc
- Checkbox to enable/disable each individual model in `<Controls />` (switch?)
- Will have to work out what the reset button should do once state is being handled in the app and not this library

kmilo9999:
The current implementation of the volume viewer doesnt facilitate the blending between voxels, because we are doing back to front ray caster. To make it easier for us, we can send multiple textures to the same model, add a prop that indicates those textures need to be blended, and add code to the shader to interpolate of them.

So we have two different cases: multi objects in the scene, and multi textures in one model. The former could apply if we want to split the scene in multiple view ports or we want a scene that simulates a "museum" of datasets  (this current PR already does it !!) , and the latter to blend volumes inside the unit cube. 

If we want to send multiple textures to the same model, we can rename the prop `channel` to `datasets` or something more meaningful, and send the array of textures on that prop. Also, add a prop that indicates the blending must be done. 

In the future we could rename rename some variables to not getting confuse with the right terminology:
The model itself is the unity cube, the information inside it (the volume) is being rendered using the png textures. We can rename `model` to something more meaningful.

PR #69, opened 2022-03-21
Fix controls state

RobertGemmaJr:
Was accidentally passing the tab index (the current model open in the `Controls`) instead of the model's index (of the `models` array) into `ColorMapControls` and `OpacityControls`. This was causing `setModels` to update the wrong model Some additional performance changes too.

- Pass correct index into `ColorMapControls` and `OpacityControls`
- Memoizes `AframeScene` (no longer re-renders when `controlsVisible` changes)
- Keep `<Controls />` on the ReactDOM at all times (prevents unnecessary re-initialization of state)

### DIFF
--- a/src/components/AframeScene/AframeScene.jsx
+++ b/src/components/AframeScene/AframeScene.jsx
@@ -1,5 +1,4 @@
-import React from "react";
-import { memo } from "react";
+import React, { memo } from "react";
 
 import "aframe";
 import "../../Aframe/arcball-camera";
--- a/src/components/AframeScene/AframeScene.jsx
+++ b/src/components/AframeScene/AframeScene.jsx
@@ -12,7 +12,6 @@ import "../../Aframe/collider-check";
 import { toAframeString } from "../../utils";
 
 function AframeScene({ models, sliders }) {
-  console.log("SCENE")
   return (
     <a-scene id="volumeViewerScene" background="color: black" embedded>
       {/* HAND */}
--- a/src/components/AframeScene/AframeScene.jsx
+++ b/src/components/AframeScene/AframeScene.jsx
@@ -1,4 +1,5 @@
 import React, { memo } from "react";
+import { isEqual } from "lodash"
 
 import "aframe";
 import "../../Aframe/arcball-camera";
@@ -11,6 +12,7 @@ import "../../Aframe/collider-check";
 import { toAframeString } from "../../utils";
 
 function AframeScene({ models, sliders }) {
+  console.log("SCENE")
   return (
     <a-scene id="volumeViewerScene" background="color: black" embedded>
       {/* HAND */}
@@ -94,4 +96,4 @@ function AframeScene({ models, sliders }) {
   );
 }
 
-export default memo(AframeScene);
+export default memo(AframeScene, (prevProps, nextProps) => isEqual(prevProps, nextProps));
--- a/src/components/AframeScene/AframeScene.jsx
+++ b/src/components/AframeScene/AframeScene.jsx
@@ -8,9 +8,10 @@ import "../../Aframe/render-2d-clipplane";
 import "../../Aframe/entity-collider-check";
 import "../../Aframe/collider-check";
 
-import { toAframeString } from "../../utils"
+import { toAframeString } from "../../utils";
 
 function AframeScene({ models, sliders }) {
+  console.log("SCENE", models);
   return (
     <a-scene id="volumeViewerScene" background="color: black" embedded>
       {/* HAND */}
--- a/src/components/AframeScene/AframeScene.jsx
+++ b/src/components/AframeScene/AframeScene.jsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { memo } from "react";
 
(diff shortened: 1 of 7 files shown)


src/components/AframeScene/AframeScene.jsx
```diff
@@ -1,4 +1,5 @@
 import React from "react";
+import { memo } from "react";
```
mcmcgrath13, line 2:
```suggestion
import React, { memo } from "react";
```

src/components/AframeScene/AframeScene.jsx
```diff
@@ -94,4 +95,4 @@ function AframeScene({ models, sliders }) {
   );
 }
 
-export default AframeScene;
+export default memo(AframeScene);
```
mcmcgrath13, line 98: no custom props equality function? I'd think this is where we'd want to use lodash's deep equal to check if the props are really different

src/components/Controls/Controls.jsx
```diff
@@ -7,35 +7,36 @@ import ColorMapControls from "../ColorMapControls";
...
 
-function Controls({ models, sliders, setModels, setSliders, reset }) {
-  // Keep track of currently open model
-  const [tabIndex, setTabIndex] = React.useState(0);
-  const handleTabsChange = (index) => {
-    setTabIndex(index);
-  };
-
-  // TODO: Each individual Tab/TabPanel should only updated when the specific model changes
+// function Controls({ models, sliders, setModels, setSliders, reset }) {
```
mcmcgrath13, line 10:
```suggestion
```

src/components/Controls/Controls.jsx
```diff
@@ -7,35 +7,36 @@ import ColorMapControls from "../ColorMapControls";
...
+  controlsVisible,
+  models,
+  sliders,
+  setModels,
+  setSliders,
+  reset,
+}) {
   return (
-    <Wrapper index={tabIndex} onChange={handleTabsChange}>
+    <Wrapper $visible={controlsVisible}>
```
mcmcgrath13, line 20: why the `$` prop? is that a convention?

src/components/Controls/Controls.jsx
```diff
@@ -7,35 +7,36 @@ import ColorMapControls from "../ColorMapControls";
...
       <StyledTabList>
         {models.map((model) => (
           <Tab key={model.name}>{model.name}</Tab>
```
mcmcgrath13, line 23: should the `handleTabsChange` move to here?

src/components/VolumeViewer/VolumeViewer.jsx
```diff
@@ -32,19 +32,20 @@ function VolumeViewer({
...
-          setModels={setModels}
-          setSliders={setSliders}
-          reset={() => {
-            setModels(buildModels(modelsProp));
-            setSliders(DEFAULT_SLIDERS);
-            setRemountKey(Math.random());
-          }}
-        />
-      )}
+      {/* {controlsVisible && ( */}
```
mcmcgrath13, line 35: delete commented out code

RobertGemmaJr: done (:

src/components/Controls/Controls.jsx
RobertGemmaJr: Reach-UI handles it internally and `tabIndex` is no longer needed so I'm just letting Reach-UI handle it

src/components/Controls/Controls.jsx
RobertGemmaJr: Reach-UI handles it internally and `tabIndex` is no longer needed so I'm just letting Reach-UI handle it

src/components/Controls/Controls.jsx
RobertGemmaJr: I was getting an warning when passing the prop to styled-components. Not entirely sure why but the `$` seems to be convention to prevent the prop from being passed to the HTML - [second anwser](https://stackoverflow.com/questions/49784294/warning-received-false-for-a-non-boolean-attribute-how-do-i-pass-a-boolean-f)

src/components/AframeScene/AframeScene.jsx
RobertGemmaJr: I can add it in but it doesn't really seem to be changing anything? I'm pretty sure the re-rendering issue was actually caused by the index bug that this PR fixes. I was wrong in thinking the props were the same because I didn't notice the bug

src/components/AframeScene/AframeScene.jsx
```diff
@@ -94,4 +95,4 @@ function AframeScene({ models, sliders }) {
   );
 }
 
-export default AframeScene;
+export default memo(AframeScene, (prevProps, nextProps) => isEqual(prevProps, nextProps));
```
mcmcgrath13, line 98:
```suggestion
export default memo(AframeScene, isEqual);
```
I think this should work

mcmcgrath13 (approved):
🌈

PR #79, opened 2022-04-11
Expose `sliders` to the user

### LINKED ISSUE #65: Pass sliders as props
Make `sliders` a prop of the VolumeViewer component, otherwise `DEFAULT_TRANSFER_FUNCTION

RobertGemmaJr:
Lets user pass `sliders` into `<VolumeViewer />` with proper validation. The value is set to `DEFAULT_SLIDERS` if the `sliders` prop is not given to the user.

The component now exports `DEFAULT_SLIDERS` and `DEFAULT_MODEL`. I figure the user might want to pick and choose what's in `DEFAULT_MODEL` in their website.

closes #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ const DEFAULT_SLIDERS = {
 
 ## DEFAULT_MODEL
 
-The `DEFAULT_MODEL` export is the default value for the `model` prop. It will be applied automatically if you do not pass `model` into `<VolumeViewer />`
+The `DEFAULT_MODEL` export contains the default values for each property of the `model` prop. It is merged with the `model` prop you pass into `<VolumeViewer />`. Note that `model.colorMap` and `model.path` are not included as they are required props.
 
 ```js
 const DEFAULT_MODEL = {
--- a/src/components/VolumeViewer/VolumeViewer.jsx
+++ b/src/components/VolumeViewer/VolumeViewer.jsx
@@ -47,7 +47,6 @@ function VolumeViewer({
     setModel(initModel);
   }, [initModel, colorMaps]);
 
-  // Always begin with DEFAULT_SLIDERS value
   const [sliders, setSliders] = useState(slidersProp);
 
   // Changing a components key will remount the entire thing


src/components/VolumeViewer/VolumeViewer.jsx
```diff
@@ -93,16 +94,51 @@ const COLOR_MAP = PropTypes.exact({
...
    * Array of color maps available in the controls.
    *  name: Common name of the color map
    *  path: Path to the color map src
    */
-  colorMaps: PropTypes.arrayOf(COLOR_MAP),
+  colorMaps: PropTypes.arrayOf(COLOR_MAP).isRequired,
 
   /** Whether or not the controls can be seen */
-  controlsVisible: PropTypes.bool,
+  controlsVisible: PropTypes.bool.isRequired,
```
mcmcgrath13, line 141: why change these to required?

src/components/VolumeViewer/VolumeViewer.jsx
```diff
@@ -161,12 +197,20 @@ VolumeViewer.propTypes = {
...
     useTransferFunction: PropTypes.bool,
-  }),
+  }).isRequired,
+
+  /* Sliders for control of clipping along the x, y, and z axes */
+  sliders: PropTypes.exact({
+    x: SLIDER,
+    y: SLIDER,
+    z: SLIDER,
+  }).isRequired,
```
mcmcgrath13, line 207: would there be a case where someone would pass in only a subset of the sliders?

src/components/VolumeViewer/VolumeViewer.jsx
RobertGemmaJr: I added it thinking it's safer to always include `isRequired` but it does seem best practice to exclude `isRequired` when the prop is given in `defaultProps`. I'll revert.

src/components/VolumeViewer/VolumeViewer.jsx
RobertGemmaJr: Perhaps? We expose `DEFAULT_SLIDERS` though so they could do

```js
sliders: {
  ...DEFAULT_SLIDERS,
  y: [/* whatever */]
```

README.md
```diff
@@ -147,6 +165,38 @@ The `ColorMaps` export is an object containing some default color maps that may
...
+const DEFAULT_SLIDERS = {
+  x: [SLIDER_RANGE.min, SLIDER_RANGE.max],
+  y: [SLIDER_RANGE.min, SLIDER_RANGE.max],
+  z: [SLIDER_RANGE.min, SLIDER_RANGE.max],
+};
+```
+
+## DEFAULT_MODEL
+
+The `DEFAULT_MODEL` export is the default value for the `model` prop. It will be applied automatically if you do not pass `model` into `<VolumeViewer />`
```
mcmcgrath13, line 183: is it applied or is it merged in with the props that are passed? - isn't the model prop generally required in some form?

src/components/VolumeViewer/VolumeViewer.jsx
```diff
@@ -47,7 +48,7 @@ function VolumeViewer({
   }, [initModel, colorMaps]);
 
   // Always begin with DEFAULT_SLIDERS value
```
mcmcgrath13, line 50: update comment

RobertGemmaJr: Good catch. I think it's safe to just delete.

mcmcgrath13 (approved):
🎚️

PR #81, opened 2022-04-12
Refractor props for blending

RobertGemmaJr:
Refractors the props as we talked about in our VR check in and adds classes for some object types

- Adds `useColorMap` prop
- Moves `slices` and `spacing` as props of `VolumeViewer`, not each individual model
- Renames `channel` as `blending` and makes it a prop of `VolumeViewer`, not each individual model
- Adds prop validation for `position`, `rotation`, `scale`, `slices`, and `spacing`
- Adds custom validation for `sliders`
- Adds `classes` folder
- Adds `Point` class - `transferFunction`  is an array of `Point`s
- Adds `ColorMap` class, default color maps are static properties
- Adds `Blending` enum as a class of static properties
- Adds `Model` classes
- Each individual model is now merged with the default properties (`Model.Default`) in the `Model` constructor
- Moves custom validation for `models` into `VolumeViewer.propTypes`, written for `Model` class and all individual properties
- Adds a `<hr />` between `<TabPanel  />` and `<ClipControls` in `Controls.jsx`
- Updates `README` accordingly

This PR includes the changes to `sliders` so merging should wait until #79 is merged. There's still some validation that needs to be done based on `useTransferFunction` and `useColorMap` but I'll work on that in another PR. 

- closes issue #80 
- closes issue #75
- closes issue #75
- closes issue #73

PR #93, opened 2022-04-18
Merge aframe blending

### LINKED ISSUE #90: Comparison function for volume update function
Need a better comparison function  between `oldData` and `data` in `volume.js` `update(oldData)`

Pick out differences between 
- blending
- model
- slices
- spacing
- sliders

Pick out which individual model changes within `models`
Pick out which model property changed within an individual model

RobertGemmaJr:
I want to get this other big branch merged in to get us back to trunk based development. Once this and `refractor-props` are in  `main` all the current changes will be merged in and I can start working on PRs as individual steps. This branch includes #81 so we'll have to wait for that to be merged in first.

- Removes `Shader.js`, `DEFAULT_MATERIAL` is moved to `constants`
- Uses `lodash` to compare `oldData` and `this.data` in `volume.js`
- Separates logic of `update` function in `volume.js` based on that comparison
- closes #90 
-


RobertGemmaJr:
Also, we're merging directly into `main` now, not a sub-branch

RobertGemmaJr:
@mirestrepo @broarr @mcmcgrath13 @kmilo9999 #81 is merged into main and pulled into this branch so the PR is ready to go!

This branch actually has a lot more to do with the update function in `volume.js` than changing the shader code. I pull the vertex and fragment shader into their own files and refactor them but don't really change the logic. Those changes are coming in the `shader-mult-uniforms` branch. That said, I removed the `#if LINEAR_FILTER` from the fragment shader (we always apply it) in a different branch but can add the change here if we think it's pertinent.

src/Aframe/fragment-shader.frag
```diff
@@ -1,114 +1,104 @@
...
-uniform vec2 viewport;
-uniform mat4 P_inv;
-uniform vec3 box_min;
-uniform vec3 box_max;
-uniform float intensity;
-varying vec3 vUV; //3D texture coordinates form vertex shader interpolated by rasterizer
+uniform mat4 clip_plane;
+uniform float dim;
+uniform float intensity;    // Artifically scale each pixel intensity
+uniform float slices;        // Number of slicess in the volumes
```
mcmcgrath13, line 12:
```suggestion
uniform float slices;       // Number of slicess in the volumes
```

src/Aframe/volume.js
```diff
@@ -133,23 +136,22 @@ AFRAME.registerComponent("volume", {
...
+        this.controllerObject.add(mesh);
+        this.grabbed = true;
+      }
+
       // Stop grabbing object
       if (this.grabbed && !triggerDown) {
         mesh.matrix.premultiply(this.controllerObject.matrixWorld);
         mesh.matrix.decompose(mesh.position, mesh.quaternion, mesh.scale);
-        this.el.object3D.add(mesh);
+        this.el.object3D.add(mesh); // Add? Shouldn't this be setObject3D?
```
mcmcgrath13, line 151: resolve this comment?

src/Aframe/volume.js
```diff
@@ -267,68 +269,99 @@ AFRAME.registerComponent("volume", {
...
+  },
+
+  updateSpacing: function () {
+    const { spacing } = this.data;
+    const uniforms = this.getMesh().material.uniforms;
+    const texture = uniforms.u_data.value; // Image
+    const dim = uniforms.dim.value;
+    const slices = uniforms.slices.value;
+
+    const volumeScale = texture
```
mcmcgrath13, line 321: this is kind of big for a ternary - maybe make an if/else?

mcmcgrath13: or get the width/height based on texture and then make the `new Vector3` once outside of control flow with those computed values

src/Aframe/volume.js
```diff
@@ -267,68 +269,99 @@ AFRAME.registerComponent("volume", {
...
   },
 
-  buildMesh: function (modelsData) {
-    this.getMesh().material = modelsData.length
-      ? modelsData[0].material
-      : new ShaderMaterial(DEFAULT_MATERIAL);
-    console.log("All models loaded", modelsData); // TEMP
+  // Blend model's into a single material and apply it to the model
+  updateMaterial: function () {
+    console.log("MODELS LOADED", this.uniforms);
```
mcmcgrath13, line 352: remove console log

src/Aframe/volume.js
```diff
@@ -267,68 +269,99 @@ AFRAME.registerComponent("volume", {
...
+
+    this.getMesh().material =
+      this.uniforms.size > 0
+        ? // TEMP - use first material
```
mcmcgrath13, line 356: resolve this comment?

src/utils/index.js
```diff
@@ -1,8 +1,34 @@
...
+  pick,
+  isArray,
+  isObject,
+  transform,
+} from "lodash";
+
+/** EXPORTS */
+
+// Custom useMemo hook for models
+function useModelsPropMemo(models) {
```
mcmcgrath13, line 15: I thought the other pr moved this into `hooks` - does a merge need to happen?

src/utils/index.js
```diff
@@ -1,8 +1,34 @@
...
+  useEffect(() => {
+    if (!noChange) previousRef.current = models;
+  });
+  return noChange ? prevModels : models;
+}
 
 // Filter model properties needed from aframe
 function getAframeModels(models) {
-  const out = models.map((model) => {
+  models = models.map((model) => {
```
mcmcgrath13, line 31: I would make this a new variable name and not override the passed variable

src/utils/index.js
```diff
@@ -29,7 +55,25 @@ function getAframeModels(models) {
 
     return aframeModel;
   });
-  return JSON.stringify(out.filter((model) => model.enabled));
+  return JSON.stringify(models.filter((model) => model.enabled));
+}
+
+// Recursively find the differences between two objects
+function deepDifference(oldObj, newObj) {
```
mcmcgrath13, line 62: there's a mix of `function` `const name = function ..` and `const name = () => ..` declarations in this file - maybe just use fat arrow functions?

src/utils/index.js
```diff
@@ -29,7 +55,25 @@ function getAframeModels(models) {
...
+  const changes = (newObj, oldObj) => {
+    let arrayIndexCounter = 0;
```
mcmcgrath13, line 64: do you have a reference link for this function? it's a bit hard to follow and could maybe use a few comments on what is happening

src/Aframe/fragment-shader.frag
```diff
@@ -1,114 +1,104 @@
...
+    data_position = camPos + t_start * ray_direction;
+    t_end = t_end - t_start;
+    t_start = 0.0;
 
     // Get t for the clipping plane and overwrite the entry point
+    // This only occurs when grabbing volume in VR
     if(clipping) {
-        vec4 p_in = clipPlane * vec4(dataPos + t_hit.x * geomDir, 1);
-        vec4 p_out = clipPlane * vec4(dataPos + t_hit.y * geomDir, 1);
+        vec4 p_in = clip_plane * vec4(data_position + t_start * ray_direction, 1);
```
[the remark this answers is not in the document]
RobertGemmaJr: They definitely seem to be some sort of convention. I've seen them around tutorials too but am not entirely sure what the stand for cc @kmilo9999

src/Aframe/vertex-shader.vert
```diff
@@ -67,17 +72,14 @@ mat4 getInverse(mat4 m) {
...
-    Result[3] = m[0] * v[0] + m[1] * v[1] + m[2] * v[2] + m[3];
-    return Result;
-}
-
 void main() {
-    mat4 MV_tmp = scale(modelViewMatrix,vec3(1,1,zScale));
-    mat4 MVP = projectionMatrix * MV_tmp;
-    gl_Position = MVP * vec4( position, 1.0 );
+    // clip space = modelViewMatrix * zScale * projectionMatrix
+    // clip space = position * scale * translation
```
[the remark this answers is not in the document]
RobertGemmaJr: Better description of the math that's going on. Meant to delete the first comment, will do so

src/Aframe/volume.js
```diff
@@ -67,19 +58,25 @@ AFRAME.registerComponent("volume", {
...
+      this.scene.canvas.height
     );
+    this.getMesh().material = new RawShaderMaterial(initMaterial);
   },
 
   update: function (oldData) {
-    const { data, usedModels, usedColorMaps } = this; // Extract read-only data
+    const { data, usedModels, usedColorMaps } = this;
+    const diffObject = deepDifference(oldData, data);
+    console.log(diffObject);
```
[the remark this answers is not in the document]
RobertGemmaJr: ✔️

src/Aframe/volume.js
RobertGemmaJr: done

src/Aframe/volume.js
RobertGemmaJr: I haven't started passing the `modelsData` array into the shader code yet. Working on it in a different branch, should I just delete the comment?

src/utils/index.js
RobertGemmaJr: Yeah, looks like it didn't get deleted during the merge. Will delete from the `utils` file

src/utils/index.js
RobertGemmaJr: Renamed to `aframeModels`

src/utils/index.js
RobertGemmaJr: I'm thinking we leave them as `function`? That seems reasonable to me since they're in a file of utility of functions. Are the arrow functions more typical in JS?

src/Aframe/volume.js
RobertGemmaJr: Looks like `add` will add a child object to the entity. In practice it's doing nothing (even if I move the code outside the `if` block and I don't think this is the desired output anyways?

@kmilo9999 Can you take a look at everything inside the `tick` function in `volume.js`? Shouldn't this be moving the object, not adding a child?

src/Aframe/volume.js
RobertGemmaJr: `texture` is undefined on the initial render. Looks like it's safe to just encapsulate everything in `if(texture) {}` because `this.updateSpacing` is called whenever the models update

src/utils/index.js
RobertGemmaJr: I added a comment with the reference as well:

https://davidwells.io/snippets/get-difference-between-two-objects-javascript

src/Aframe/volume.js
RobertGemmaJr: This does get me thinking though. Should we enforce that `texture.image.width` and `texture.image.height` are the same for every `model` in `models`? I can only image this will become an issue when we blend the textures together. 

I imagine tiff to png outputs the same sized images but just in case... or is it safe to just let it error out if that becomes a problem?

src/Aframe/fragment-shader.frag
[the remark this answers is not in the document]
mirestrepo: I think `t` is a translation amount in the direction if the ray and that starts. And `p` are the point in 3-d for entry and exit of the clipping plain. These look like point/line equations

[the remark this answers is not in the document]
mcmcgrath13: could we add a comment to that effect?

src/Aframe/volume.js
mcmcgrath13: you can just cite the issue where it will be resolved

mirestrepo (commented):
Aside from the comments from @mcmcgrath13 , I don't spot major red flags from my side, but I think @kmilo9999 needs to give a look as he'll be able to comment more on the logic/math!

src/Aframe/fragment-shader.frag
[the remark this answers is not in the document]
RobertGemmaJr: I noted `t_` as a translation vector and `p_` as a position vector

src/Aframe/volume.js
RobertGemmaJr: Okay! It's all set

mcmcgrath13 (approved):
Looks good to me from a javascript/general code quality perspective! I defer to @kmilo9999 on the graphics content

kmilo9999:
This PR doesnt have the texture array loader on the aframe component or the shaders. The javascript code looks good.

Issue #94, opened 2022-04-19
`setModel` function in `Controls.jsx` 


RobertGemmaJr:
Pass `setModel` instead of `setModels` into each component inside `<Controls>`. Would call ass `setModel(newModel, idx)`

PR #96, opened 2022-05-03
Set model func

RobertGemmaJr:
Creates a `setModel` function that only updates a single model in the `models` state array. 

`setModel` takes an index and the change applied to that model

closes issue #94


src/components/Controls/ColorMapControls.jsx
```diff
@@ -22,16 +22,14 @@ function ColorMapControls({ model, modelIdx, setModels }) {
           aria-labelledby="ColorMap dropdown"
           value={model.colorMap.name}
           onChange={(newColorMapName) =>
-            setModels((models) => [
```
mcmcgrath13, line 25: I'd consider making these even more specific to the component, so this prop would be `setColorMap` and then on the parent the function would be something like `setColorMap={(colorMap) => setModel({colorMap}, index)`

The `modelIndex` is really the business of the parent and I don't think the child needs to know about it at all.

RobertGemmaJr: It's causing some weird issues with `usecallback`. When I make a custom `setTransferFunction` it's causing the <TransferFunctionControls />` to re-run on every render

RobertGemmaJr: Seems like I need to make each individual function (`setColorMap`, `setTransferFunction`) a `useCallback` function but `setModel` is already a callback function

RobertGemmaJr: Pulling `setTransferFunction` out of the useEffect dependency array (in `TransferFunctionControls`) seems to have done the trip. eslint is mad but I disabled the warning, is it safe to do so?

src/components/Controls/TransferFunctionControls.jsx
```diff
@@ -122,18 +121,16 @@ function TransferFunctionControls({
...
+    setTransferFunction(
+      canvasPoints.map((p) => {
+        return {
+          x: scaleTransferFunctionToCanvasX.invert(p.x),
+          y: scaleTransferFunctionToCanvasY.invert(p.y),
+        };
+      })
     );
-  }, [canvasPoints, modelIdx, pointHovering, pointDragging, setModel]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
mcmcgrath13, line 132: I think that's probably fine in this case to ignore the `setTransferFunction` prop here - @broarr could you sanity check?

mcmcgrath13 (approved):
pending @broarr 's sanity check

RobertGemmaJr:
I chatted with Brad in person, he said it's all set! Should definitely start working on moving to context though

PR #98, opened 2022-05-04
Always use the linear filter

RobertGemmaJr:
`LINEAR_FILTER` is defined as true at the top of the fragment shader and never changes. I see no reason to keep the variable but let me know if y'all disagree


mcmcgrath13 (commented):
I defer to people who know graphics on this one (cc @eldu )

eldu (approved):
Looks good to me ! Seems like it was hardcoded anyway and this simplifies the code !

kmilo9999 (commented):
:thumbsup: :thumbsup:

PR #99, opened 2022-05-04
Upgrade shader code

RobertGemmaJr:
Upgrades shader code to use OpenGL `#version 300 es`. The desktop volume viewer uses `#version 330 core` but that is not available in WebGl. The closest version seems to be 300 es.

### DIFF
--- a/src/Aframe/vertex-shader.vert
+++ b/src/Aframe/vertex-shader.vert
@@ -5,8 +5,8 @@ out vec3 vUV;       // Coordinates of the texture
 out vec3 camPos;    // Coordinates of the camera
 
 uniform float zScale;
-uniform mat4 modelViewMatrix;   // Position of model in world space (WebGL)
-uniform mat4 projectionMatrix;  // Position of the model in clip space (WebGL)
+uniform mat4 modelViewMatrix;   // world -> camera space transformation matrix (WebGL) 
+uniform mat4 projectionMatrix;  // camera -> clip space transformation matrix (WebGL)
 
 
 mat4 scale(mat4 m, vec3 v) {
--- a/src/Aframe/vertex-shader.vert
+++ b/src/Aframe/vertex-shader.vert
@@ -1,11 +1,13 @@
-uniform float zScale;
-varying  vec3 vUV;      // 3D texture coordinates for texture lookup in the fragment shader
-varying  vec3 camPos;   // Position of the camera
+#version 300 es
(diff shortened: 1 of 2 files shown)


mcmcgrath13 (commented):
I defer to the graphics experts on this one

src/Aframe/vertex-shader.vert
```diff
@@ -1,11 +1,13 @@
...
+
+in vec3 position;   // Position vector of the vertex (WebGL)
+out vec3 vUV;       // Coordinates of the texture
+out vec3 camPos;    // Coordinates of the camera
+
 uniform float zScale;
-varying  vec3 vUV;      // 3D texture coordinates for texture lookup in the fragment shader
-varying  vec3 camPos;   // Position of the camera
+uniform mat4 modelViewMatrix;   // Position of model in world space (WebGL)
+uniform mat4 projectionMatrix;  // Position of the model in clip space (WebGL)
```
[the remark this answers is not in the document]
RobertGemmaJr: Oh okay that totally makes sense! I updated the comments

src/Aframe/vertex-shader.vert
```diff
@@ -1,11 +1,13 @@
...
+
+in vec3 position;   // Position vector of the vertex (WebGL)
+out vec3 vUV;       // Coordinates of the texture
+out vec3 camPos;    // Coordinates of the camera
+
 uniform float zScale;
-varying  vec3 vUV;      // 3D texture coordinates for texture lookup in the fragment shader
-varying  vec3 camPos;   // Position of the camera
+uniform mat4 modelViewMatrix;   // world -> camera space transformation matrix (WebGL) 
+uniform mat4 projectionMatrix;  // camera -> clip space transformation matrix (WebGL)
```
kmilo9999, line 9: I thought Webgl (threejs )  already [defined this matrices](https://threejs.org/docs/#api/en/renderers/webgl/WebGLProgram) . Do you need to redefine them to go up on glsl ?

RobertGemmaJr: It defines them with `THREE.ShaderMaterial` but I'm using `THREE.RawShaderMaterial`. My linter was getting very mad when I was using `ShaderMaterial` so I decided to switch. THREE still handles all the setting/updating of the matrices

kmilo9999 (approved):
It looks good!

PR #104, opened 2022-05-11
Utilize THREE caching

RobertGemmaJr:
three.js has it's own caching system when it comes to loading files: https://threejs.org/docs/#api/en/loaders/Cache

- Refactor `loadColorMap` to utilize `THREE.ImageLoader`
- Remove `usedColorMaps` and `usedModels` - let THREE manage caching

### DIFF
--- a/src/Aframe/volume.js
+++ b/src/Aframe/volume.js
@@ -63,6 +63,8 @@ AFRAME.registerComponent("volume", {
       this.scene.canvas.height
     );
     this.getMesh().material = initMaterial;
+
+    console.log("INIT", Object.keys(THREE.Cache.files).length);
   },
 
   update: function (oldData) {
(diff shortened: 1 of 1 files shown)


kmilo9999:
What file needs the line `THREE.Cache.enabled= true;`   ?

mcmcgrath13 (approved):
Nice find! do you notice any improvements with this caching?

RobertGemmaJr:
> What file needs the line `THREE.Cache.enabled= true;` ?

It's enabled automatically when we use `THREE.TextureLoader`, `THREE.DataTexture`, etc

> Nice find! do you notice any improvements with this caching?

Nothing that I could tell just by eye but it definitely decreases the amount of data we're passing around which may end up being our bigger issue

PR #116, opened 2022-06-03
Feat: Average Blending

### LINKED ISSUE #112: Add more blending types
- Add
- Multiple

What others are standard types of blending?

### LINKED ISSUE #115: Add average blending
Blend fragment =  sum of fragments colors/ number of samples.
We might have to clamp the final color in case gets > 1.0

RobertGemmaJr:
Adds logic for blending the average of each model.

The GLSL `mix()` function mixes based on a percentage (0 - 1) of the two inputs. The ratio of the first input (`v_sample.a`) to the total (`v_sample.a + m_sample.a`) yields that percentage. 

- closes #112 
- closes #115


broarr (approved):

src/aframe/fragment-shader.frag
```diff
@@ -136,34 +138,35 @@ void main() {
...
-                        model_sample = sample_model(model_structs[i], data_position);
-                        volume_sample = mix(volume_sample, model_sample, mix_factor);
+                        m_sample = sample_model(model_structs[i], data_position);
+                        v_sample = mix(v_sample, m_sample, mix_factor);
                     }
                 }
                 #pragma unroll_loop_end
             }
-        } else break; // array is "empty", leave vFragColor transparent
+        } else break; // No models in use, leave transparent
```
kmilo9999, line 166: place  `model_sample.rgb *= model.intensity;`  here. So the value of the pixel only gets increased once per model and not once per sample.

src/aframe/fragment-shader.frag
```diff
@@ -136,34 +138,35 @@ void main() {
 
     // Starting from the entry point, march the ray through the volume and sample it
     vec4 vFragColor = vec4(0);
-    vec4 model_sample, volume_sample;
+    vec4 m_sample, v_sample;
```
kmilo9999, line 141: `m_sample ` is only used inside the internal for loop. you can move it below `if(blending != 0)`

kmilo9999 (commented):
👍  keep the good work !

kmilo9999 (changes requested):
Adding comments to my review

example/src/App.js
```diff
@@ -98,7 +98,7 @@ function App() {
   const VV = (
     <StyledVolumeViewer
       controlsVisible={controlsVisible}
-      blending={Blending.Max}
```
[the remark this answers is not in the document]
RobertGemmaJr: Will do!

src/aframe/fragment-shader.frag
RobertGemmaJr: Actually yeah I refactored a little more to calculate the mix factor correctly and it doesn't actually make a difference? Should each model have its own intensity or should it be a property of the volume itself that's calculated after all of the blending is done?

src/aframe/fragment-shader.frag
RobertGemmaJr: Moved it

src/aframe/fragment-shader.frag
kmilo9999: it should be calculated after all the blending is done. let's apply the max intensity of the models.

RobertGemmaJr: Okay it should be all set!

RobertGemmaJr:
@eldu can you take a look at the shader code while Camilo is on vacation? Thank you!

RobertGemmaJr:
Consolidating with other Blending changes to #131

eldu (commented):
Added in some comments and questions! Looks good to me though !

Issue #111, opened 2022-06-02
Move coordinate math our of `sampleAs3DTexture`


RobertGemmaJr:
Performance improvement - `sampleAs3DTexture` is called multiple times for the same coordinates. Should calculate in `create_volume` loop instead

RobertGemmaJr:
`create_volume` has been deleted, now in main

PR #119, opened 2022-06-03
Refactor coordinates calculation

RobertGemmaJr:
- Removes the `sampleAs3DModel` function
- Pulls the coordinate calculation out of a for loop (was doing the same calculation over again)
- closesissue #111


kmilo9999:
Probably renaming the variables is also a good idea.
So, data_position comes form the vertex shader and it is a 3D coordinate that is in world coordinates. It's not in the 0-1 normalized space.  So the `floor` function is just clamping to the nearest integer, which it wont be 0 all the times.
Are you getting rendering issues ?

RobertGemmaJr:
> Are you getting rendering issues ?

Nothing that I can notice

RobertGemmaJr:
@kmilo9999 Can you give this a review when you get the chance? I renamed some of the variables and added comments to make things a little more clear

kmilo9999 (approved):
Seems good to me,  Good work!

PR #125, opened 2022-06-14
Match README Example

RobertGemmaJr:
The example code in README is now the same as the deployed example app

- Adds a button to swap whether or not the models are enabled
- Slight change to the first model's initial transfer function
- Moves all of the JSX code into the `return` in `App.js`

### DIFF
--- a/src/components/AframeScene/AframeScene.jsx
+++ b/src/components/AframeScene/AframeScene.jsx
@@ -25,7 +25,7 @@ function AframeScene({
   useEffect(() => {
     const handler = (error) => {
       const messages = error.detail.map((e) => {
-        console.error(e)
+        console.error(e);
         if (e.cause) return e.message + ". " + e.cause.message;
         else return e.message;
       });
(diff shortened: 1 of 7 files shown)


src/components/AframeScene/AframeScene.jsx
```diff
@@ -25,7 +25,7 @@ function AframeScene({
   useEffect(() => {
     const handler = (error) => {
       const messages = error.detail.map((e) => {
-        console.error(e)
+        console.error(e);
```
jashlu, line 28: curious why do some lines require a semi colon while others don't? like in `VolumeViewer.jsx` line 50. Everything looks good!

jashlu (approved):

src/components/AframeScene/AframeScene.jsx
RobertGemmaJr: I'll just run an `npm run lint` again, that should fix the issue

PR #126, opened 2022-06-14
fix: error when no enabled models

RobertGemmaJr:
Fixes a validation error that was occurring when no enabled models were passed to `aframe` and refactors the code into its own function.


jashlu (approved):
Look's good! 🎉

PR #133, opened 2022-06-23
Fix: Fix colormap import to work on new release

RobertGemmaJr:
I originally was importing all the colormap images using `require.context` but that was causing errors when I went to release a new version of the package. Now using hard-coded imports to get all of the images.


eldu (approved):
Looks good to me ! 🚀

Issue #134, opened 2022-07-05
check texture sizes should be optional (on by default)


kmilo9999:
Blending two datasets of different number of slices throws an error due to texture sizes restrictions.
This should be optional in the library  in case we find this cases.  related to  #113

RobertGemmaJr:
I'm not sure what you want me to implement. Right now they always have to be the same size. We would have to calculate the start and end position for each individual model again in order for this to work, which is fine if that's what you're looking for

kmilo9999:
It should be possible to blend multiple datasets with different sizes. Right now, it throws an error when the volumes have a different width and height. Could you remove this check and/or make it optional?

PR #135, opened 2022-07-13
Use max texture size for zScale

RobertGemmaJr:
- Removes `validateModelTextureSizes` function
- Uses maximum width and height to determine `zScale`
- closes issue #134


kmilo9999 (commented):
It's mostly javascript code. It's look good for me.
@broarr @jashlu  could you quickly check?

PR #136, opened 2022-07-22
fix blend intensity

kmilo9999:
remove intensity calculation based on maximum value per group of voxels. changed to individual value per voxel


RobertGemmaJr (changes requested):
Looks great! I would lint the tabbing to be safe.

Make sure you do a version bump in `package.json`, otherwise the new release won't work.

package.json
```diff
@@ -29,9 +29,14 @@
     "@reach/listbox": "^0.16.2",
     "@reach/tabs": "^0.16.4",
     "aframe": "^1.2.0",
+    "bootstrap": "^4.6",
```
RobertGemmaJr, line 32: Only the version bump should have changed, not sure what's going on with all the dependencies
```suggestion
```

package.json
```diff
@@ -29,9 +29,14 @@
...
     "d3-scale": "^4.0.2",
     "lodash": "^4.17.21",
-    "rc-slider": "^10.0.0"
+    "rc-slider": "^10.0.0",
+    "react-bootstrap": "^1.6.3",
+    "sass": "^1.44.0",
+    "styled-components": "^5.0.0",
+    "web-vitals": "^1.0.1"
```
RobertGemmaJr, line 39:
```suggestion
```

package.json
```diff
@@ -52,6 +57,7 @@
     "gh-pages": "^3.2.3",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.4.1",
+    "prettier-plugin-glsl": "^0.0.4",
```
RobertGemmaJr, line 60: Can you change this to a dev dependency?

```console
npm uninstall prettier-plugin-glsl
```

```console
npm install prettier-plugin-glsl --save-dev

File: src/aframe/fragment-shader.frag [hunk 4]
```diff
@@ -99,81 +99,74 @@ void main() {
...
-
+        if(vFragColor.a>=.95)break;
+        
         // Advance point
-        data_position += ray_direction * step_size;
+        data_position+=ray_direction*step_size;
     }
-    fragColor = vFragColor;
+    fragColor=vFragColor;
 }
```
RobertGemmaJr, line 172: Just adding a newline to the end of the file
```suggestion
}

```

RobertGemmaJr (commented):
Looks great! Something messed up with the dependencies on your local repo though. I added the suggested changes to fix them and then it's good to go!

package.json
kmilo9999: I removed it

package.json
kmilo9999: I removed them

src/aframe/fragment-shader.frag
kmilo9999: I added the new line at the end of the file. The spaces between the equals sign were already there.

RobertGemmaJr (approved):
Perfect! I'll merge and create the new release

PR #140, opened 2022-08-23
Fix blend color using multiple transfer functions

kmilo9999:
- Added blue and red color scale maps. 
- Blending data per channel. 
- Switch to non linear color blending


src/images/colormaps/index.js
```diff
@@ -50,6 +52,8 @@ const colormaps = {
   Thermal: { name: "thermal", path: thermal },
   Topo: { name: "topo", path: topo },
   Turbid: { name: "turbid", path: turbid },
+  Red_scale: { name: "red_scale", path: red_scale },
+  Blue_scale: { name: "blue_scale", path: blue_scale },
```
RobertGemmaJr, line 56: I would just leave as red and blue because the names aren't taken yet. Also, could you put them in alphabetical order for completions sake?
```suggestion
  Red: { name: "red", path: red_scale },
  Blue: { name: "blue", path: blue_scale },
```

src/aframe/fragment-shader.frag
```diff
@@ -136,24 +137,44 @@ void main() {
         
         // Sample and mix models into a single volume
         
+        int has_passed=0;
         #pragma unroll_loop_start
-        for(int i=0;i<4;i++){
+        for(int i=0;i<3;i++){
```
RobertGemmaJr, line 142: Good catch!

src/aframe/fragment-shader.frag
```diff
@@ -136,24 +137,44 @@ void main() {
...
+                    else if(blending==2){
+                        // mix uses a percentage - get ratio of the alphas
+                        mix_factor=v_sample.a/(v_sample.a+m_sample.a);
+                    }
+                    
+                    // Combine colors
+                    // non linear interpolation of colors
+                    vec3 temp=v_sample.rgb*v_sample.rgb;
+                    vec3 temp2=m_sample.rgb*m_sample.rgb;
+                    vec3 mix_color=sqrt(mix(temp,temp2,.5));
```
RobertGemmaJr, line 173: I think it's safe to combine these into a one-liner instead of creating the temporary variables. Why are you squaring the values only to take the square root?

RobertGemmaJr: Also (correct me if I'm wrong) shouldn't this be another if based on blending? Like `else if(blending==3){}`

I would assume that if we're choosing the min/max/average pixel value (e.g. for RIDDC) then we don't want to do this interpolation between the colors. It'll take a little more effort but it's not too difficult to add another value to the blending enum

src/aframe/fragment-shader.frag
```diff
@@ -136,24 +137,44 @@ void main() {
...
-                    // mix uses a percentage - get ratio of the alphas
-                    mix_factor=v_sample.a/(v_sample.a+m_sample.a);
+                m_sample[i]*=model_structs[i].intensity;
+                // make sure the color is within transparency range.
+                m_sample[i]=clamp(m_sample[i],0.,1.);
+                if(has_passed==0)
+                {
+                    //preserve rgb value from the transfer function when rendering a single model
+                    has_passed=1;
+                    v_sample=m_sample;
```
RobertGemmaJr, line 156: Can you just use `i` here? Or is that one of those shader code issues where it won't compile when you use the counter
```suggestion
                if(i==0)
                {
                    //preserve rgb value from the transfer function when rendering a single model
                    v_sample=m_sample;
```

RobertGemmaJr (commented):
Is there a reason all the spaces in the shader code disappeared? Is that a linter thing? The tabbing looks a little off in some places as well.

Otherwise it looks pretty good! Just a couple questions just for my own sanity

src/aframe/fragment-shader.frag
```diff
@@ -136,24 +137,44 @@ void main() {
...
+                        mix_factor=v_sample.a/(v_sample.a+m_sample.a);
+                    }
+                    
+                    // Combine colors
+                    // non linear interpolation of colors
+                    vec3 temp=v_sample.rgb*v_sample.rgb;
+                    vec3 temp2=m_sample.rgb*m_sample.rgb;
+                    vec3 mix_color=sqrt(mix(temp,temp2,.5));
+                    
+                    // Result alpha value is the factor calculated by the bleding mode.
```
eldu, line 175: typo
```suggestion
                    // Result alpha value is the factor calculated by the blending mode.
```

src/aframe/fragment-shader.frag
```diff
@@ -136,24 +137,44 @@ void main() {
...
+                if(has_passed==0)
+                {
+                    //preserve rgb value from the transfer function when rendering a single model
+                    has_passed=1;
+                    v_sample=m_sample;
+                }else
+                {
+                    
+                    // Do data and non linear color blending
+                    // Calculate the alpha mix factor (0: Max, 1: Min, 2: Average)
```
eldu, line 161: Would ratio be a better term for what's happening here? Since in lines 165/166 that's what you're referring to?

For an average I'd expect something like `(v_sample.a + m_sample.a) / 2`

I don't think I fully understand what `mix_factor=v_sample.a/(v_sample.a+m_sample.a)` is trying to do. If `v_sample.a = 0.5` and `m_sample.a = 0.5` would you want a `mix_factor` of `1`?

```suggestion
                    // Calculate the alpha mix factor (0: Max, 1: Min, 2: Ratio)
```

eldu (commented):
I added in some comments! I'll leave the approving up to Rob ;D

src/images/colormaps/index.js
kmilo9999: I think `Red` will imply that is only a hard a color. I used the postfix `scale` indicating that there some kind of transition. I'm open to use another prefix/postfix.

src/aframe/fragment-shader.frag
kmilo9999: This part of the logic is just to check if it is rendering one model or blending multiple models. `i` is the index of the cycle.

src/aframe/fragment-shader.frag
kmilo9999: You are very right.  @RobertGemmaJr , do you want me to rework that formula to the one `(v_sample.a + m_sample.a) / 2` ? or leave it and change it on latter PR?

src/aframe/fragment-shader.frag
RobertGemmaJr: That's handled in `if(model_structs[i].use){` in line 143. If only one model is being rendered then it will only enter the loop once. I think it makes sense to initialize `v_sample` to the first model though, that logic definitely makes sense. Great catch!

So `if(i==0) { v_sample=m_sample; }` and then we never enter this code block again (because of `if(model_structs[i].use)` or we enter the `else`

src/aframe/fragment-shader.frag
RobertGemmaJr: If the colors in the RIDDC models don't need to be blended together then do they need to use this interpolation? I guess that's the ultimate question. If they still need it then it's perfect but if they don't then we need to rework the logic

src/aframe/fragment-shader.frag
RobertGemmaJr: No, mix factor is a percentile - 0 would only take `v_sample`, 0.5 is the correct value for `v_sample == m_sample` and 1 would only take `m_sample`.

Also, because we're inside the loop we can't just do `(v_sample.a + m_sample.a) / 2` because `v_sample` is already an average of multiple models if that makes sense. It's a little hard to explain but I'll try to throw in a screenshot explaining the math

src/images/colormaps/index.js
RobertGemmaJr: Okay sounds good! I would name them as `RedScale` and `BlueScale` to keep the naming scheme then

src/aframe/fragment-shader.frag
RobertGemmaJr: Okay you changed how the final `v_sample` is calculated so let me think through the math a little bit more

image

RobertGemmaJr: Yeah @kmilo9999 correct me if I'm wrong but as far as I can tell this non-linear interpolation of the colors has nothing to do with the other three blending types. That or (and?) the original blending for the RIDDC viewer has always been wrong. 

**Blending.Max: blending=0**
```frag
mix_factor= m_sample.a >= v_sample.a ? 1 : 0

// Then we mix
v_sample=mix(v_sample, m_sample, mix_factor)
```

**Blending.Min: blending=1**
```frag
mix_factor= m_sample.a <= v_sample.a ? 1 : 0

// Then we mix
v_sample=mix(v_sample, m_sample, mix_factor)
```

**Blending.Average: blending=2**

_Let me know if you still want me to spell out the math - when v_sample.a = m_sample.a then mix_factor = 0.5_
```frag
mix_factor=v_sample.a/(v_sample.a+m_sample.a);

// Then we mix
v_sample=mix(v_sample, m_sample, mix_factor)
```

**Blending.Quadratic: blending=3**

```frag
// The value in `mix_factor` (current implementation) is really just the final alpha value
// I would think it's best to use the average alpha value from above?
alpha= ? 

// Combine colors
// non linear interpolation of colors
vec3 temp=v_sample.rgb*v_sample.rgb;
vec3 temp2=m_sample.rgb*m_sample.rgb;
vec3 mix_color=sqrt(mix(temp,temp2,.5)); // Mix is done here - mix_factor is 0.5

// Result alpha value is the factor calculated by the bleding mode.
v_sample=vec4(mix_color,mix_factor);
```

RobertGemmaJr: I can help with the refactor. Depending on the deadline your up against we should definitely get this right

src/aframe/fragment-shader.frag
kmilo9999: Done!

src/images/colormaps/index.js
kmilo9999: Done.

src/aframe/fragment-shader.frag
kmilo9999: There is no requirement on blending RIDDC data. Those datasets are on different scales (Temp and Salt). At one point, we thought about a multi-variable transfer function, but the discussion never went too far for start working on it. The data on the microscopy viewer is on the same scale, so the blending was a requirement.
The data blends properly, that's why the grayscale view of the microscopy viewer you can see multiple models/structures combined on a same volume. The problem was combining the colors from the transfer functions (the purple color issue).

src/aframe/fragment-shader.frag
kmilo9999: What if I load 3 models but I want to see only the second model? Or blend just the last two models? Initializing `v_sample` to first model won't work. Maybe initializing it to the first available model (could be the second one) makes more sense. But, I believe this kind of logic can be (somehow) moved to the javascript code. The shader should be responsible only for the pixel logic.
You might already know this, but the` #pragma unroll_loop_start` makes the compiler to generate the same code inside of the cycle N times. So, the cycle really doesnt exist. That's why I used the `has_passed` variable. I'm open to new suggestions here. Although, we could create an issue with this and change it on a future PR.

src/aframe/fragment-shader.frag
kmilo9999: You are very right on the average calculation. Technically, it's an average, but the formula has a name different to Average, correct? I can rename it  thought.

> Yeah @kmilo9999 correct me if I'm wrong but as far as I can tell this non-linear interpolation of the colors has nothing to do with the other three blending types.

Correct. I dont think we need a new type of blending. We are always doing this non-linear interpolation when we are blending multiple models. When rendering only one, we dont need the quadratic formula. We just go straight for the color in the transfer function color map. Hope that makes sense.

src/aframe/fragment-shader.frag
RobertGemmaJr: Yeah it might be best for a different PR but I'm pretty sure the array sent to the fragment shader always has the first enabled model (the toggle in the controls) at `i=0` to combat that issue. I'd have to look through the code again to make sure that's the case but it should be fairly simple to get

kmilo9999: It does not. But of you tell me that the first model is always the first enabled model, then we can work it out the way you expect it.

RobertGemmaJr: I would have to take another look through the code to make sure I'm remembering correctly, I think it's best to leave it as is for now. We can work it out in another PR

src/aframe/fragment-shader.frag
RobertGemmaJr: I would just rename `mix_factor` to `alpha` because that's really what the value is

RobertGemmaJr: I'm wondering if it'll be easier to have an entirely different fragment shader file when only using a single model. Different PR obviously but might be nice to have the logic in `volume.js` instead of the shader code.

src/aframe/fragment-shader.frag
```diff
@@ -136,24 +137,42 @@ void main() {
...
         
+        int has_passed=0;
         #pragma unroll_loop_start
-        for(int i=0;i<4;i++){
+        for(int i=0;i<3;i++){
             if(model_structs[i].use){
                 // Sample model and mix in to volume
+                // m_sample.rgb is equals to the value found in the color map look up table
+                // m_sample.a is equals to the alpha value obtained from the transfer function widget
                 vec4 m_sample=sample_model(model_structs[i],start,end,mix_position);
```
RobertGemmaJr, line 147:
```suggestion
                // m_sample.rgb is  the value found in the color map look up table
                // m_sample.a is the value obtained from the transfer function widget
                vec4 m_sample=sample_model(model_structs[i],start,end,mix_position);
```

RobertGemmaJr (approved):
Quick wording change and change `mix_factor` to `alpha` and I think we're ready to go!

src/aframe/fragment-shader.frag
kmilo9999: On a C++ program is possible.  Probably on Javascript as well. Just have to work out the logic. The issue is that the voxel logic would be same for both shaders, it would be ideal to separate it on a separate file and include it on both shaders. However GLSL  doesnt support the `#include` directive and we will be repeating code either way.

PR #143, opened 2022-08-31
ref: Move useColorMap outside of the individual model

### LINKED ISSUE #137: Allow for greyscale transfer function to be in-sync when using multiple channels
See [this issue](https://github.com/brown-ccv/microscopy-viewer/issues/12)

RobertGemmaJr:
`useColorMap` is now a prop of the component and not the individual models. Every model becomes grayscale when `!useColorMap`. Added a checkbox in the controls and handle them in state

- closes #137

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,6 @@ VolumeViewer.propTypes = {
    *      Array of {x: <val>, y: <val>} coordinates.
    *      Each coordinate in transferFunction must be between (0, 0) and (1,1)
    *    useTransferFunction: Flag to apply a transfer function to the model
-   *    useColorMap: Flag to apply a color map to the model
    */
   models: MODEL,
 
@@ -92,6 +91,9 @@ VolumeViewer.propTypes = {
     y: SLIDER,
     z: SLIDER,
   }),
+
+  /** Enable the color map controls (otherwise use grayscale) */
+  useColorMaps: PropTypes.bool,
 };
 ```
 
@@ -110,7 +112,8 @@ VolumeViewer.defaultProps = {
     x: [0, 1],
     y: [0, 1],
     z: [0, 1],
-  };,
+  },
+  useColorMaps: true,
 };
 ```
 
@@ -184,7 +187,6 @@ It will be merged with every `model` in the `models` array. Note, however, that 
     { x: 1, y: 1 },
   ],
   useTransferFunction: true,
-  useColorMap: true,
 };
 ```
 
@@ -246,7 +248,6 @@ function App() {
             { x: 1, y: 1 },
           ],
           useTransferFunction: useTransferFunction,
-          useColorMap: useColorMap,
         },
         {
           name: "Temperature",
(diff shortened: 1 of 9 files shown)


kmilo9999:
@RobertGemmaJr  Can you place the "Use Color Map" checkbox at the top of the options tab? It could be hard to know that the option exists.

User case: When the  "Use Color Map"  is unchecked, then modify the clip planes, and then click on the `reset` button. The datasets is color mapped, but the "Use Color Map" is still unchecked.

kmilo9999 (changes requested):
@RobertGemmaJr Can you place the "Use Color Map" checkbox at the top of the options tab? It could be hard to know that the option exists.

User case: When the "Use Color Map" is unchecked, then modify the clip planes, and then click on the reset button. The datasets is color mapped, but the "Use Color Map" is still unchecked.

RobertGemmaJr:
> Can you place the "Use Color Map" checkbox at the top of the options tab? It could be hard to know that the option exists.

The only issue is differentiating which controls are for an individual model (color map, transfer function) and which are for the whole volume (useColorMap and the sliders). I could flip them so the "global" controls are on top?

> User case: When the "Use Color Map" is unchecked, then modify the clip planes, and then click on the reset button. The datasets is color mapped, but the "Use Color Map" is still unchecked.

Looks like this is a bug, I'll fix it in a second

kmilo9999 (changes requested):
I see your point. Let's leave the check box on the place you proposed it. Why dont we change the logic on that ? Rename the label to "**Use gray scale color map**" and is  `unchecked` by default. When the user checks the box, then grayscale color map is applied.

User case: I just added a control point to the transfer function and if I click on the reset button, the point remains. I thought it was a cache problem, but I also tested it on a new chrome session and the button is still there. I think is better to fix this problem asap and dont wait for a new PR.

RobertGemmaJr:
> Rename the label to "**Use gray scale color map**" and is `unchecked` by default. When the user checks the box, then grayscale color map is applied.

We talked about what to do a fair bit at the GSDC meeting today. I'm not sure this logic really make sense, isn't grayscale literally the absence of color?

> User case: I just added a control point to the transfer function and if I click on the reset button, the point remains.

I haven't been able to recreate this problem? Whatever is added to the transfer function from the controls disappears when the reset button is clicked

RobertGemmaJr:
@kmilo9999 @mirestrepo @eldu I've made a different version (#144) where the global controls sit on top of the others. Let me know what you think. Just looking at the layout for right now, I can clean up the style of 144 if that's preferred

kmilo9999:
> We talked about what to do a fair bit at the GSDC meeting today. I'm not sure this logic really make sense, isn't grayscale literally the absence of color?

Well, the absence of color is black. The data itself produce a grayscale image ( or volume ). We map the data value to a RGB color located in the texture. My opinion is that "color mapping" is a technical concept.  The users that have seen the app ( at least Rachel and Jeff ) talk more about grayscale that color maps. You can make the argument that the title on the tabs says "Color Map", thus makes sense to call the checkbox "Use color map".  What do you think?

> > User case: I just added a control point to the transfer function and if I click on the reset button, the point remains.
> 
> I haven't been able to recreate this problem? Whatever is added to the transfer function from the controls disappears when the reset button is clicked

You are right, I didnt notice you have a pre-defined transfer function burnt in the example app.

RobertGemmaJr:
Closing as stale

PR #144, opened 2022-09-02
useColorMap above

RobertGemmaJr:
Alternative version of #143 where the global controls sit on top of the tabs. If this version looks better I can clean up the background/styling of it


kmilo9999 (commented):
I dont like this new layout.

PR #145, opened 2023-06-05
Fix VR Mode

kmilo9999:
- Upgrade to aframe 1.4.0
- Removed unused code to reposition volume mesh after exiting vr mode
- Added code to recognize vr controller input from openxr
- Added logic to reset volume mesh position after exiting vr mode
- Added sphere-point test to simulate grab interaction between vr controller and volume mesh

### DIFF
--- a/example/package-lock.json
+++ b/example/package-lock.json
@@ -15,12 +15,12 @@
       }
     },
     "..": {
-      "version": "1.1.0",
+      "version": "1.1.4",
       "license": "MIT",
       "dependencies": {
         "@reach/listbox": "^0.16.2",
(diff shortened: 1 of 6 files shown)


RobertGemmaJr (approved):
Great! Does entering and exiting VR now always start everything on the origin? Love to see it! 🚀

kmilo9999:
> Great! Does entering and exiting VR now always start everything on the origin? Love to see it! 🚀

Hey Robert. Yes, moving the mesh around in VR - Mode affects its world position in the web scene. It needs to be place in the center in the scene in order to work with the arc-ball camera.

PR #148, opened 2023-06-12
Fix float precision for OpenglES

kmilo9999:
- Adding `highp` qualifier at computing ray casting - box intersection. The lack of floating precision was generating artifacts in oculus browser and VR mode (Meta moved to OpenXR implementation)
- Updating bay volume datasets


src/aframe/fragment-shader.frag
```diff
@@ -54,6 +54,9 @@ vec4 sample_model(ModelStruct model, vec2 start_position, vec2 end_position, flo
     model_sample.a = max(model_sample.r, max(model_sample.g, model_sample.b));
     if(model_sample.a < 0.20) model_sample.a *= 0.1;
     
+    // uncomment next line and comment next return statement to render basic volume cube
+    // return vec4(1.0,0.0,0.0,1.0);
```
RobertGemmaJr, line 58: I'm not sure what you mean here by the comment?

RobertGemmaJr (approved):
Just a clarification question but looks good! 🚀

src/aframe/fragment-shader.frag
kmilo9999: If you uncomment line 58 and comment 61 to 64, it would render the cube volume. It's just for debugging purposes in case I want to check new shading code affects the ray casting algorithm.

